PR #11, opened 2015-01-13
Removing "Failed to loads bookmarks"

theCoolkal:
This warning in Zathura is a nuisance for a number of reasons:
* Warning appears under perfectly normal conditions as you may not have made any bookmarks or may not use them
* Using Zathura as a PDF viewer with vim-latex can cause text to be written over the terminal when it shows this unnecessary warning as it repeatedly appears whenever the PDF is recompiled
* There are other warning and error messages for when the bookmarks file really can't be read
* `Failed to loads bookmarks` has an annoying typo

### DIFF
--- a/zathura/zathura.c
+++ b/zathura/zathura.c
@@ -828,7 +828,7 @@ document_open(zathura_t* zathura, const char* path, const char* password,
   /* bookmarks */
   if (zathura->database != NULL) {
     if (zathura_bookmarks_load(zathura, file_path) == false) {
-      girara_warning("Failed to loads bookmarks.");
+      girara_debug("Failed to load bookmarks.");
     }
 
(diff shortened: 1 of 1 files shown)


sebastinas:
Could you instead please demote the warning to debug, i.e. call `girara_debug`, and fix the typo? What is the purpose of the merge commit?

theCoolkal:
Okay. How about this?

I did a merge because I accidentally did it in my master branch instead of develop.

sebastinas:
Could you rebase it on top of our develop branch to get rid of the merge commit? Or should I cherry-pick the other two commits?

theCoolkal:
Okay. How about now?

sebastinas:
Thank you! Merged.

PR #12, opened 2015-02-06
Fix dual page render breakge of adjust_window

xlz:
When the page is set to adjust width to the window, toggling dual page mode would break rendering because
sc_toggle_page_mode() did not update dimensions.

To reproduce,
1. Open a file
2. Press "s", switch to fit-to-width mode
3. Press "d", switch to dual page mode
4. Broken pages are shown.

This patch resolves https://bugs.pwmt.org/issue446.


mlq:
Thank you!

PR #13, opened 2015-03-23
Include ${DESTDIR} in pkgconfig includedir.

pitti:
(See also pwmt/girara#3)

Signed-off-by: Philipp Ittershagen <pit@shgn.de>


sebastinas:
No. That's not what DESTDIR is for.

PR #16, opened 2015-04-28
replaced deprecated gdk_cursor_new()

michelesr:
While compiling zathura from git i noticed a deprecation warning for the [gdk_cursor_new()](https://developer.gnome.org/gdk3/stable/gdk3-Cursors.html#gdk-cursor-new) function so i replaced it with [gdk_cursor_new_for_display()](https://developer.gnome.org/gdk3/stable/gdk3-Cursors.html#gdk-cursor-new-for-display).


sebastinas:
Thanks, merged.

PR #17, opened 2015-05-08
Feature: Shortcut for exiting the program in index mode

haggl:
Simply defines the key 'q' as a shortcut for exiting zathura in index mode.
As a prerequisite I had to fix .gitignore, as git wouldn't let me stage anything in the "zathura"-folder.


sebastinas:
Thank you. Merged.

PR #18, opened 2015-09-12
Automagically set first-page-column in sc_toggle_page_mode

haggl:
When switching between single- and double-page layout, first-page-column is set
to 2 (double-page) or reset to 1 automatically. In the author's opinion, this
is more intuitive behaviour - especially when viewing books.


sebastinas:
I'm not really happy with this approach. It may be acceptable if the user opens a document for the first time, but not afterwards. Maybe we should change first-page-column to have a default values plus values for different number of colums. Something like `default value:value for two columns:value for three columns:...` where empty or omitted values are replaced by the default value.

haggl:
Oh, I didn't realize that `first-page-column` is saved separately for each document.

I'll look into it *after* finishing my Diploma thesis. ;-)

haggl:
I have opened a new pull-request: https://github.com/pwmt/zathura/pull/20

PR #19, opened 2015-10-01
Add feature to show ~ instead of $HOME in title

sid-kap:
This allows users to add "set window-title-home-tilde true" to their zathurarc so that, in the window title, $HOME is replaced with a "~".

Please let me know if I should rename functions or fix my style anywhere, or if you have a suggestion for a better name for this feature.

### DIFF
--- a/zathura/config.c
+++ b/zathura/config.c
@@ -224,6 +224,8 @@ config_load_default(zathura_t* zathura)
   bool_value = false;
   girara_setting_add(gsession, "window-title-basename",  &bool_value,  BOOLEAN, false, _("Use basename of the file in the window title"), NULL, NULL);
   bool_value = false;
+  girara_setting_add(gsession, "window-title-home-tilde",  &bool_value,  BOOLEAN, false, _("Use ~ instead of $HOME in the filename in the window title"), NULL, NULL);
+  bool_value = false;
   girara_setting_add(gsession, "window-title-page",      &bool_value,  BOOLEAN, false, _("Display the page number in the window title"), NULL, NULL);
   bool_value = false;
   girara_setting_add(gsession, "statusbar-basename",     &bool_value,  BOOLEAN, false, _("Use basename of the file in the statusbar"), NULL, NULL);
--- a/zathura/config.c
+++ b/zathura/config.c
@@ -229,6 +229,8 @@ config_load_default(zathura_t* zathura)
   girara_setting_add(gsession, "window-title-page",      &bool_value,  BOOLEAN, false, _("Display the page number in the window title"), NULL, NULL);
   bool_value = false;
   girara_setting_add(gsession, "statusbar-basename",     &bool_value,  BOOLEAN, false, _("Use basename of the file in the statusbar"), NULL, NULL);
+  bool_value = false;
+  girara_setting_add(gsession, "statusbar-home-tilde",  &bool_value,  BOOLEAN, false, _("Use ~ instead of $HOME in the filename in the statusbar"), NULL, NULL);
   bool_value = true;
   girara_setting_add(gsession, "synctex",                &bool_value,  BOOLEAN, false, _("Enable synctex support"), NULL, NULL);
   string_value = "";
--- a/zathura/zathura.c
+++ b/zathura/zathura.c
@@ -547,13 +547,24 @@ document_info_open(gpointer data)
 }
 
 const char*
-get_window_title_filename(zathura_t* zathura, const char* file_path)
+get_formatted_filename(zathura_t* zathura, const char* file_path, bool statusbar)
 {
   bool basename_only = false;
-  girara_setting_get(zathura->ui.session, "window-title-basename", &basename_only);
+  if (statusbar) {
+    girara_setting_get(zathura->ui.session, "window-title-basename", &basename_only);
+  } else {
+    girara_setting_get(zathura->ui.session, "statusbar-basename", &basename_only);
+  }
+
   if (basename_only == false) {
+
     bool home_tilde = false;
-    girara_setting_get(zathura->ui.session, "window-title-home-tilde", &home_tilde);
+    if (statusbar) {
+      girara_setting_get(zathura->ui.session, "statusbar-home-tilde", &home_tilde);
+    } else {
+      girara_setting_get(zathura->ui.session, "window-title-home-tilde", &home_tilde);
+    }
+
     if (home_tilde) {
       char *home = getenv("HOME");
       size_t home_len = home ? strlen(home) : 0;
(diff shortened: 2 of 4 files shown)


zathura/config.c
```diff
@@ -224,6 +224,8 @@ config_load_default(zathura_t* zathura)
   bool_value = false;
   girara_setting_add(gsession, "window-title-basename",  &bool_value,  BOOLEAN, false, _("Use basename of the file in the window title"), NULL, NULL);
   bool_value = false;
+  girara_setting_add(gsession, "window-title-home-tilde",  &bool_value,  BOOLEAN, false, _("Use ~ instead of $HOME in the filename in the window title"), NULL, NULL);
+  bool_value = false;
```
sebastinas: Please also document the new setting in the manpage.

zathura/zathura.c
```diff
@@ -546,6 +546,38 @@ document_info_open(gpointer data)
...
+      int file_path_len = file_path ? strlen(file_path) : 0;
+
+      if (home_len > 1 && strncmp(home, file_path, home_len) == 0 && (!file_path[home_len] || file_path[home_len] == '/')) {
+        // Length should be total length of path - length of $HOME + 1 for '~' + 1 for '\0'
+        int tlen = file_path_len - home_len + 2;
+        char *tdir = malloc(sizeof(char) * tlen);
+        strncpy(tdir + 1, file_path + home_len, tlen);
+        tdir[0] = '~';
+        tdir[tlen - 1] = '\0';
+        return tdir;
```
sebastinas: This leaks the memory allocated in line 565. Also, we use the glib based malloc  and string functions, please continue using them here too. Otherwise it's a pain to keep track where we have to use free or g_free.

zathura/zathura.c
```diff
@@ -546,6 +546,38 @@ document_info_open(gpointer data)
...
+{
+  bool basename_only = false;
+  girara_setting_get(zathura->ui.session, "window-title-basename", &basename_only);
+  if (basename_only == false) {
+    bool home_tilde = false;
+    girara_setting_get(zathura->ui.session, "window-title-home-tilde", &home_tilde);
+    if (home_tilde) {
+      char *home = getenv("HOME");
+      int home_len = home ? strlen(home) : 0;
+      int file_path_len = file_path ? strlen(file_path) : 0;
```
sebastinas: The type of strlen's return values is size_t, not int.

zathura/zathura.c
sid-kap: Sure, I can change the malloc to g_try_malloc.

How should I avoid this memory leak? I need to return the value of tdir, so I can't g_free it in this method.

zathura/zathura.c
sid-kap: Fixed

zathura/zathura.c
```diff
@@ -546,6 +546,49 @@ document_info_open(gpointer data)
...
+
+    if (home_tilde) {
+      char *home = getenv("HOME");
+      size_t home_len = home ? strlen(home) : 0;
+      size_t file_path_len = file_path ? strlen(file_path) : 0;
+
+      if (home_len > 1 && strncmp(home, file_path, home_len) == 0 && (!file_path[home_len] || file_path[home_len] == '/')) {
+        // Length should be total length of path - length of $HOME + 1 for '~' + 1 for '\0'
+        int tlen = file_path_len - home_len + 2;
+        char *tdir = g_try_malloc(sizeof(char) * tlen);
```
sid-kap: Can you explain why this might cause a memory leak? Sorry, I'm not very experienced in C.

zathura/zathura.c
```diff
@@ -546,6 +546,49 @@ document_info_open(gpointer data)
...
+
+    bool home_tilde = false;
+    if (statusbar) {
+      girara_setting_get(zathura->ui.session, "statusbar-home-tilde", &home_tilde);
+    } else {
+      girara_setting_get(zathura->ui.session, "window-title-home-tilde", &home_tilde);
+    }
+
+    if (home_tilde) {
+      char *home = getenv("HOME");
```
sebastinas: Please use girara_get_home_directory instead. It also checks passwd.

zathura/zathura.c
sebastinas: The string is passed to girara_statusbar_item_set_text and girara_set_window_title but is never freed and so the memory is leaked. This function needs to either return a string that needs to be passed to g_free or a string that does not need to be freed at all, but not a mixture of both. If strings need to freed with g_free, call sites need to take care of that.

zathura/zathura.c
sid-kap: According to girara documentation (https://pwmt.org/projects/girara/doxygen/utils_8c.html#a71314474e74a8da69cafaab9155d0f12), $HOME is preferred to girara_get_home_directory. So maybe we should continue using $HOME.

Anyway, if you think it's better to use `girara_get_home_directory`, I can do that. I have a question though: girara_get_home_directory takes a `const char* user` as a parameter. Where can I get this info? It looks like girara_fix_path tries to find the pattern `~/(.*)/` in the path to find the username. This seems kind of sloppy. Is there a better way?

sebastinas: The documentation says: the function returns $HOME if it is set and fallsback to other means if not.

`user` is also documented on the page you linked: the user or `NULL` to get the home directory of the current user.

sid-kap: Apologies for not reading it carefully. Thanks!

sid-kap:
I believe I've fixed the memory leak. Is this correct?

zathura/zathura.c
```diff
@@ -546,6 +546,57 @@ document_info_open(gpointer data)
...
+        return path;
+      }
+    } else {
+      char* path = g_try_malloc(sizeof(char) * strlen(file_path));
+      strncpy(path, file_path, file_path_len);
+      return path;
+    }
+  } else {
+    const char* basename = zathura_document_get_basename(zathura->document);
+    char* basename_copy = g_try_malloc(sizeof(char) * strlen(basename));
```
sebastinas: You can just use g_strdup here.

sebastinas:
Looks better now, yes.

zathura/zathura.c
sid-kap: Thanks for the suggestion! I've fixed this in 3 places.

sid-kap:
Awesome. I would love to get more feedback, in particular about the naming of my settings (window-title-home-tilde and statusbar-home-tilde), and get this merged in!

zathura/zathura.c
```diff
@@ -546,6 +546,51 @@ document_info_open(gpointer data)
...
+      girara_setting_get(zathura->ui.session, "window-title-home-tilde", &home_tilde);
+    }
+
+    size_t file_path_len = file_path ? strlen(file_path) : 0;
+
+    if (home_tilde) {
+      char *home = girara_get_home_directory(NULL);
+      size_t home_len = home ? strlen(home) : 0;
+
+      if (home_len > 1 && strncmp(home, file_path, home_len) == 0 && (!file_path[home_len] || file_path[home_len] == '/')) {
```
sebastinas: This also needs a `file_path_len >= home_len` check.

zathura/zathura.c
```diff
@@ -546,6 +546,51 @@ document_info_open(gpointer data)
...
+        // Length should be total length of path - length of $HOME + 1 for '~' + 1 for '\0'
+        int tlen = file_path_len - home_len + 2;
```
sebastinas: Wrong type. But I'd just replace this whole block with something like `g_strdup_printf("~/%s", file_path + home_len + 1)`.

sid-kap:
I added the additional check that you suggested, and incorporated more glib string functions to make the code simpler.

zathura/zathura.c
```diff
@@ -546,6 +546,55 @@ document_info_open(gpointer data)
...
+
+      if (home_len > 1
+          && file_path_len >= home_len
+          && g_str_has_prefix(file_path, home)
+          && (!file_path[home_len] || file_path[home_len] == '/')) {
+
+        size_t remaining_len = file_path_len - home_len;
+
+        GString* remaining_path = g_string_new_len(&file_path[home_len], remaining_len);
+        char* tilde_path = g_strdup_printf("~%s", remaining_path->str);
```
sebastinas: Shouldn't `file_path + home_len` not be enough here?

sebastinas:
Thanks you for the patch! I've merged it with a small memory leak fix.

PR #20, opened 2015-11-07
Feature/toggle page mode improvement

haggl:
Extends `first-page-column` to a string containing a list of first page column values:
```
<single page layout>:<double page layout>[: ...]
```
The default value for the list is `1:2`, so in double-page layout the renderer places the first page in the second column.

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -243,8 +243,13 @@ cb_page_layout_value_changed(girara_session_t* session, const char* name, girara
   unsigned int pages_per_row = 1;
   girara_setting_get(session, "pages-per-row", &pages_per_row);
 
-  unsigned int first_page_column = 1;
-  girara_setting_get(session, "first-page-column", &first_page_column);
+  /* get list of first_page_column settings */
+  char* first_page_column_list = NULL;
+  girara_setting_get(session, "first-page-column", &first_page_column_list);
+
+  /* find value for first_page_column */
+  unsigned int first_page_column = find_first_page_column(first_page_column_list, pages_per_row);
+  g_free(first_page_column_list);
 
   unsigned int page_padding = 1;
   girara_setting_get(zathura->ui.session, "page-padding", &page_padding);
--- a/zathura/database-sqlite.c
+++ b/zathura/database-sqlite.c
@@ -5,6 +5,7 @@
 #include <girara/datastructures.h>
 #include <girara/input-history.h>
 #include <string.h>
+#include <strings.h>
 
 #include "database-sqlite.h"
 #include "utils.h"
@@ -17,6 +18,7 @@ G_DEFINE_TYPE_WITH_CODE(ZathuraSQLDatabase, zathura_sqldatabase, G_TYPE_OBJECT,
                         G_IMPLEMENT_INTERFACE(GIRARA_TYPE_INPUT_HISTORY_IO, io_interface_init))
 
 static bool           check_column(sqlite3* session, const char* table, const char* col, bool* result);
+static bool           check_column_type(sqlite3* session, const char* table, const char* col, const char* type, bool* result);
 static void           sqlite_finalize(GObject* object);
 static bool           sqlite_add_bookmark(zathura_database_t* db, const char* file, zathura_bookmark_t* bookmark);
 static bool           sqlite_remove_bookmark(zathura_database_t* db, const char* file, const char* id);
@@ -148,7 +150,7 @@ sqlite_db_init(ZathuraSQLDatabase* db, const char* path)
     "scale FLOAT,"
     "rotation INTEGER,"
     "pages_per_row INTEGER,"
-    "first_page_column INTEGER,"
+    "first_page_column TEXT,"
     "position_x FLOAT,"
     "position_y FLOAT,"
     "time TIMESTAMP"
@@ -176,9 +178,9 @@ sqlite_db_init(ZathuraSQLDatabase* db, const char* path)
 
   /* update fileinfo table (part 2) */
   static const char SQL_FILEINFO_ALTER2[] =
-    "ALTER TABLE fileinfo ADD COLUMN first_page_column INTEGER;";
+    "ALTER TABLE fileinfo ADD COLUMN first_page_column TEXT;";
 
-  /* update fileinfo table (part 2) */
+  /* update fileinfo table (part 3) */
   static const char SQL_FILEINFO_ALTER3[] =
     "ALTER TABLE fileinfo ADD COLUMN time TIMESTAMP;";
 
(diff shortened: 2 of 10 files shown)


zathura/utils.c
```diff
@@ -257,3 +257,26 @@ GdkAtom* get_selection(zathura_t* zathura)
 
   return selection;
 }
+
+unsigned int find_first_page_column(const char* first_page_column_list,
+                                    const unsigned int pages_per_row) {
+  /* copy settings list */
+  char* settings = malloc(strlen(first_page_column_list));
+  strcpy(settings, first_page_column_list);
```
sebastinas: The buffer is not large enough. Please just use `g_strdup` here. Also, the buffer is never freed.

zathura/utils.c
```diff
@@ -257,3 +257,26 @@ GdkAtom* get_selection(zathura_t* zathura)
...
+
+  /* save first setting as default value for omitted settings */
+  unsigned int first_page_column = atoi(strtok(settings, ":"));
```
sebastinas: Please use the reentrant version of `strtok`, `strtok_r`, or just use `g_strsplit` as we do elsewhere. You do not have to copy the string if you use the latter.

zathura/zathura.c
```diff
@@ -651,13 +652,14 @@ document_open(zathura_t* zathura, const char* path, const char* password,
   }
 
   /* read history file */
+  char first_page_column_list_default[] = "1:2";
```
sebastinas: This buffer is certainly not large enough for anything longer.

zathura/callbacks.c
```diff
@@ -243,8 +243,12 @@ cb_page_layout_value_changed(girara_session_t* session, const char* name, girara
   unsigned int pages_per_row = 1;
   girara_setting_get(session, "pages-per-row", &pages_per_row);
 
-  unsigned int first_page_column = 1;
-  girara_setting_get(session, "first-page-column", &first_page_column);
+  /* get list of first_page_column settings */
+  char* first_page_column_list;
```
sebastinas: Please initialize with ` NULL`.

zathura/database-sqlite.c
```diff
@@ -631,14 +632,14 @@ sqlite_get_fileinfo(zathura_database_t* db, const char* file,
...
-  file_info->pages_per_row     = sqlite3_column_int(stmt, 4);
-  file_info->first_page_column = sqlite3_column_int(stmt, 5);
-  file_info->position_x        = sqlite3_column_double(stmt, 6);
-  file_info->position_y        = sqlite3_column_double(stmt, 7);
+  file_info->current_page           = sqlite3_column_int(stmt, 0);
+  file_info->page_offset            = sqlite3_column_int(stmt, 1);
+  file_info->scale                  = sqlite3_column_double(stmt, 2);
+  file_info->rotation               = sqlite3_column_int(stmt, 3);
+  file_info->pages_per_row          = sqlite3_column_int(stmt, 4);
+  file_info->first_page_column_list = (char*) sqlite3_column_text(stmt, 5);
```
sebastinas: This needs a `g_strdup`.

zathura/shortcuts.c
```diff
@@ -1308,14 +1308,14 @@ sc_toggle_presentation(girara_session_t* session, girara_argument_t*
...
-    /* setm ode */
+    /* set mode */
     girara_mode_set(session, zathura->modes.normal);
   } else if (old_mode == zathura->modes.normal) {
     /* backup pages per row */
     girara_setting_get(session, "pages-per-row", &pages_per_row);
 
     /* backup first page column */
-    girara_setting_get(session, "first-page-column", &first_page_column);
+    girara_setting_get(session, "first-page-column", first_page_column_list);
```
sebastinas: This won't work. It expects a `char**`.

zathura/zathura.c
```diff
@@ -1072,14 +1078,14 @@ document_close(zathura_t* zathura, bool keep_monitor)
...
-  zathura_fileinfo_t file_info = { 0, 0, 1, 0, 1, 1, 0, 0 };
+  zathura_fileinfo_t file_info = { 0, 0, 1, 0, 1, "1:2", 0, 0 };
   file_info.current_page = zathura_document_get_current_page_number(zathura->document);
   file_info.page_offset  = zathura_document_get_page_offset(zathura->document);
   file_info.scale        = zathura_document_get_scale(zathura->document);
   file_info.rotation     = zathura_document_get_rotation(zathura->document);
 
   girara_setting_get(zathura->ui.session, "pages-per-row", &(file_info.pages_per_row));
-  girara_setting_get(zathura->ui.session, "first-page-column", &(file_info.first_page_column));
+  girara_setting_get(zathura->ui.session, "first-page-column", &(file_info.first_page_column_list));
```
sebastinas: This leaks the memory allocated by `girara_setting_get`.

zathura/zathura.c
haggl: Hmm.. True, but it is not really a buffer, but merely a default value. The only case it is modified is when `file_info.page_column_list` is an empty string, in which case the same string "1:2" is written by the statement
```C
girara_setting_get(zathura->ui.session, "first-page-column", &first_page_column_list);
```
However, I must admit that I don't quite understand, where that value comes from. ;-)

What am I missing?

sebastinas: I may have mixed `first_page_column_list` and `first_page_column_list_default`, so that might not be an issue after all.

zathura/database-sqlite.c
```diff
@@ -148,7 +148,7 @@ sqlite_db_init(ZathuraSQLDatabase* db, const char* path)
     "scale FLOAT,"
     "rotation INTEGER,"
     "pages_per_row INTEGER,"
-    "first_page_column INTEGER,"
+    "first_page_column TEXT,"
```
sebastinas: This change needs code to alter old databases similar to the other `ALTER` statements we have to migrate old database layouts.

zathura/shortcuts.c
```diff
@@ -1308,14 +1308,14 @@ sc_toggle_presentation(girara_session_t* session, girara_argument_t*
...
-    /* setm ode */
+    /* set mode */
     girara_mode_set(session, zathura->modes.normal);
   } else if (old_mode == zathura->modes.normal) {
     /* backup pages per row */
     girara_setting_get(session, "pages-per-row", &pages_per_row);
 
     /* backup first page column */
-    girara_setting_get(session, "first-page-column", &first_page_column);
+    girara_setting_get(session, "first-page-column", &first_page_column_list);
```
sebastinas: This causes a memory leak as soon as this code is executed twice.

zathura/database-sqlite.c
```diff
@@ -178,7 +179,7 @@ sqlite_db_init(ZathuraSQLDatabase* db, const char* path)
   static const char SQL_FILEINFO_ALTER2[] =
     "ALTER TABLE fileinfo ADD COLUMN first_page_column INTEGER;";
```
sebastinas: This should be changed to `STRING` so that we can skip the extra step at the end.

zathura/database-sqlite.c
```diff
@@ -242,6 +243,26 @@ sqlite_db_init(ZathuraSQLDatabase* db, const char* path)
     }
   }
 
+  /* check existing tables for correct column types */
+  ret1 = check_column_type(session, "fileinfo", "time", "TEXT", &res1);
+
+  if (ret1 == true && res1 == false) {
+    girara_debug("old database table layout detected; updating ...");
+
+    char SQL_FILEINFO_ALTER4[525] = "BEGIN TRANSACTION;"
```
sebastinas: Please don't hardcode string sizes. We will certainly forget to update the size if one of the statements changes in the future. Either calculate it from the three parts or generate the final string dynamically.

zathura/utils.c
```diff
@@ -257,3 +257,31 @@ GdkAtom* get_selection(zathura_t* zathura)
 
   return selection;
 }
+
+unsigned int find_first_page_column(const char* first_page_column_list,
+                                    const unsigned int pages_per_row) {
```
sebastinas: We typically use
```
return_type
function_name(arguments)
{
...
}
```

haggl: Oops. I'll move the opening brace to a separate line.

zathura/database-sqlite.c
```diff
@@ -242,6 +244,37 @@ sqlite_db_init(ZathuraSQLDatabase* db, const char* path)
...
   }
 
+  /* check existing tables for correct column types */
+  ret1 = check_column_type(session, "fileinfo", "first_page_column", "TEXT", &res1);
+
+  if (ret1 == true && res1 == false) {
+    girara_debug("old database table layout detected; updating ...");
+
+    /* prepare transaction */
+    char tx_begin[] = "BEGIN TRANSACTION;"
```
sebastinas: `static const` as the others.

zathura/database-sqlite.c
```diff
@@ -242,6 +244,37 @@ sqlite_db_init(ZathuraSQLDatabase* db, const char* path)
...
+                    "COMMIT;";
+
+    /* calculate requred buffer size */
+    size_t tx_buffer_size = strlen(tx_begin);
+    tx_buffer_size += strlen(SQL_FILEINFO_INIT);
+    tx_buffer_size += strlen(tx_end);
+    ++tx_buffer_size;
+
+    /* assemble transaction */
+    char transaction[tx_buffer_size];
```
sebastinas: `char transaction[sizeof(tx_begin) + sizeof(SQL_FILEINFO_INIT) + sizeof(tx_end)] = { '\0' }` and remove the `bzero`.

sebastinas: I'll change that myself and merge it.

sebastinas:
This PR has been merged. Thank you!

PR #22, opened 2016-02-05
Fixed compilation without Synctex

Louisvh:
Compilation of init_zathura(..) would fail if synctex was not available,
as the variable synctex_editor was passed but never declared outside of
the WITH_SYNCTEX ifdefs. That's fixed now.

### DIFF
--- a/zathura/main.c
+++ b/zathura/main.c
@@ -89,7 +89,7 @@ init_zathura(const char* config_dir, const char* data_dir,
     char* synctex_editor, Window embed)
 #else
     char* synctex_editor)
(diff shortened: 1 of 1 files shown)


Louisvh:
I just realised there is a develop branch, so I apologize for creating a PR against master. However, this is a bug that completely prevents people who don't have/want SyncTeX from compiling the program and the change is trivial to review. It's unlikely to result in a non-trivial merge conflict down the line, so I do think merging is a good idea. You call though.

sebastinas:
We only merge to master on a new release. But a release with this fix should be available soon(TM).

PR #25, opened 2016-02-15
Add neovim synctex support via msgpack

kmarius:
python workaround


alerque:
Is there any setup on the Neo**vim** side needed to use this? Why did you close the PR?

kmarius:
Sorry, I'm not used to working with github, didn't mean to PR.
In fact, a python script is needed, because there does not seem to be 
working a msgpack-rpc implementation in C.

PR #27, opened 2016-10-14
 Make page mode toggling predictable + clicking a page set the current page number

fxkr:
Previously, toggling page mode would jump to near page, but usually not the one your were looking at before. Now it's more predictable.

Making clicking a page set the current page number allows fast visual navigation through the 'd'ouble page shortcut in combination with a high pages-per-row setting. Very, very handy when writing a masters thesis.


fxkr:
Should have been against develop, not master.

PR #28, opened 2016-10-14
Make page mode toggling predictable + clicking a page set the current page number

fxkr:
Previously, toggling page mode would jump to near page, but usually not the one your were looking at before. Now it's more predictable.

Making clicking a page set the current page number allows fast visual navigation through the 'd'ouble page shortcut in combination with a high pages-per-row setting. Very, very handy when writing a masters thesis.

(The previous pull request was retracted because it was against master, not develop. My mistake.)

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -634,11 +634,19 @@ void
 cb_page_widget_scaled_button_release(ZathuraPage* page_widget, GdkEventButton* event,
     void* data)
 {
+  zathura_t* zathura = data;
+
+  zathura_page_t* page = zathura_page_widget_get_page(page_widget);
+
+  /* set page number (but don't scroll there. it was clicked on, so it's visible) */
+  if (event->button == 1) {
+    zathura_document_set_current_page_number(zathura->document, zathura_page_get_index(page));
+    refresh_view(zathura);
+  }
+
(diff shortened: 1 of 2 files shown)


zathura/page-widget.c
```diff
@@ -800,6 +800,10 @@ cb_zathura_page_widget_button_press_event(GtkWidget* widget, GdkEventButton* but
       priv->mouse.selection.y2 = -1;
     }
 
+    /* set page number (but don't scroll there. it was clicked on, so it's visible) */
+    zathura_document_set_current_page_number(priv->zathura->document, zathura_page_get_index(priv->page));
+    refresh_view(priv->zathura);
```
sebastinas: While this certainly will work, I'd prefer to have the page widget emit a signal and have it handled somewhere else. Maybe add a callback that contains that code, emit a signal here and connect the callback and the signal in `zathura.c`.

fxkr:
I'm really sorry for not getting back to this sooner.

Thanks for the review; now that you point it out I agree that's cleaner.

I rebased both commits onto develop and updated the second commit as you suggested. There was already a button release event, and I don't really have any opinion about whether the page number should be set on mouse down or up, so I just reused that.

sebastinas:
Thank you, merged!

PR #29, opened 2017-02-23
add vertical centering option

dlahoti:
This is a pretty simple change – basically, when I'm TeXing something I want to be able to scroll page by page while staying slightly zoomed in so I don't have to waste screen space on the margins.


sebastinas:
Merged, thank you.

PR #30, opened 2017-04-06
elaborate manpage description of hotkey: ^r

zaz:
I only found this feature by accident; never would have found it by searching the manpage.  This PR should fix that.


sebastinas:
Thank you, merged.

PR #31, opened 2017-05-21
add continuous history save

dlahoti:
Some of my zathura sessions are pretty long-running, e.g. when I have a textbook open, often more likely to close because of something crashing or because I run out of battery than because I intentionally close them, which means I lose my page. This fixes that problem, and hopefully other people will find this option useful as well.


sebastinas:
Thank you, merged.

PR #32, opened 2017-11-08
Check for symlink 

du0ngpv:
When i create symlink of my pdf then open it , zathura open it from page 1 , not last position


sebastinas:
Please create the pull request against develop.

alerque:
Most of the commits showing up in this PR are because it's being submitted to the wrong branch. You should be able to change what branch the PR is targeted to merge into. After that if all the extraneous commits don't go away you should rebase your branch locally against `develop` and force push it again to update this PR.

du0ngpv:
Sorry i dont know how to use github :( i just want to make your software better

alerque:
@du0ngpv At the top of this pull request, there is a line that looks like this:

> du0ngpv wants to merge 61 commits into master from develop 

You need to edit this PR to change the "master" in that line to be "develop" as well. This is the target branch you are submitting to. Once you change that this should update and look much more reasonable because only your new code will be submitted to the version that it was based on.

PR #33, opened 2018-01-02
Cases for vertical_center were swapped

qwence:
Setting vertical-center = false (the default) actually produced vertical centering, and vice versa. I just swapped the cases.


sebastinas:
Thank you, merged!

PR #34, opened 2018-01-31
Fix Edit signal arguments

yitzchak:
The signature for the Edit signal indicates that line and column will be passed but it appears that x and y are being passed instead. This PR fixes this and passes line and column as intended.


sebastinas:
Thank you. Merged in to our develop branch.

PR #36, opened 2020-03-12
Fix "Synctex Highlights stack, completely obscure text" (https://git.pwmt.org/pwmt/zathura/issues/27)

W95Psp:
This PR fixes https://git.pwmt.org/pwmt/zathura/issues/27

The problem is that synctex gives loads of nested rectangles to zathura when pointing at a zone with nested Tex groups.

This PR takes these (sometimes overlapping) rectangles and flatten them. The function `flatten_rectangles` extracts the vectrices of the rectangles, and then `cut_rectangle` cuts each rectangle x and y wise relatively to every vectrices. This approach produces too much rectangles, but it's kind of simple, and it's working. So I'm not that sure whether this should be merged ; a better algo should be implemented, but I was too lazy to do so in C.

### DIFF
--- a/zathura/synctex.c
+++ b/zathura/synctex.c
@@ -203,7 +203,7 @@ synctex_rectangles_from_position(const char* filename, const char* input_file,
   }
 
   girara_list_t* hitlist_flat = flatten_rectangles(hitlist);
-  /* girara_list_free(hitlist); */
+  girara_list_free(hitlist);
   
   return hitlist_flat;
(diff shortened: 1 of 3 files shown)


sebastinas:
Thanks, merged with some modifications to fit the coding style.

W95Psp:
Nice! (And yeah, thanks for the modifications ; I'm really not a C guy ^^)

PR #39, opened 2021-08-14
Add copy file path mapping

elig0n:


sebastinas:
Thanks, applied in 8369e79.

PR #42, opened 2021-12-01
Fixes to account for variable page size

ViktorWalter:
Multiple PDF files I have dealt with were originally displayed with two issues if they had non-uniform page sizes (e.g. large resolution scan of the first page of a scanned book):
1: They were by default zoomed out significantly, such that multiple pages could fit into the view on the default zoom level.
2: They were displayed with large black margins, such that the actual page was small with respect to the view, approximately in the size it has relative to the resolution of the largest page.
This pull request provides changes derived from @ailrk 's fix of the latter issue, expanded to also fix the former.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ compile_commands.json
 *.log
 .ycm_extra_conf.py
 _*/
+.ccls-cache/
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,3 @@ compile_commands.json
 *.log
 .ycm_extra_conf.py
 _*/
-.ccls-cache/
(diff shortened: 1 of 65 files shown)


ViktorWalter:
I'll look into it.

ViktorWalter:
Ok, I think I've unified the worst offenders. Is it now legible?

sebastinas:
There are still plenty of unrelated changes. The PR seems to revert a bunch of completely unrelated stuff.

ViktorWalter:
I have merged the language files from upstram. Currently, ` git diff ` only lists the relevant files.

lambdcalculus:
Any news on this? It's still ready to merge.

sebastinas:
@ViktorWalter Thanks again … finally merged.

PR #44, opened 2022-12-10
Add GitHub build action

dylanmtaylor:


alerque:
I'm not sure how much sense a GH action makes for a project primarily hosted on GitLab. You know this repo is just a mirror right? A GitLab CI config might be welcome but this seems out of place.

dylanmtaylor:
> I'm not sure how much sense a GH action makes for a project primarily hosted on GitLab. You know this repo is just a mirror right? A GitLab CI config might be welcome but this seems out of place.

I can make one for GitLab as well if that's desirable

alerque:
I would suggest opening an issue or MR on the GitLab for the upstream to get the maintainer's attention, not here.

Also just a side note when you build anything with Arch Linux it's advisable to start from the base-devel image instead of base. It's bigger but will get cached as a Docker layer and save a lot of network and CI runner time not having to install all the basic deps for building stuff

sebastinas:
Gitlab CI is already being used: https://git.pwmt.org/pwmt/zathura/-/pipelines

PR #570, opened 2024-05-13
Add support for loongarch

wuruilong01:
Some architectures lack fstat system calls, such as loongarch


sebastinas:
This change was applied earlier today.

PR #575, opened 2024-05-25
DRAFT: Add landlock support

valoq:
This adds landlock support to the sandbox feature.

While the seccomp filter should already prevent writing to files, this adds another safeguard to prevent file modifications.

Todo:
- fix dependency check in meson (check for kernel version)
- cleanup landlock.c (remove unused functions)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -79,6 +79,7 @@ flags = cc.get_supported_arguments(flags)
 additional_sources = []
 synctex = dependency('synctex', version: '>=1.19', required: get_option('synctex'))
 seccomp = dependency('libseccomp', required: get_option('seccomp'))
+#landlock = dependency('linux', version: '>=5.13', required: get_option('landlock'))
 
 if synctex.found()
   build_dependencies += synctex
@@ -92,6 +93,11 @@ if seccomp.found()
   build_dependencies += seccomp
   defines += '-DWITH_SECCOMP'
   additional_sources += files('zathura/seccomp-filters.c')
+endif
+
+if get_option('landlock').auto()
+  defines += '-DWITH_LANDLOCK'
+  additional_sources += files('zathura/landlock.c')
 endif
 
 # generate version header file
--- a/meson.build
+++ b/meson.build
@@ -79,7 +79,7 @@ flags = cc.get_supported_arguments(flags)
 additional_sources = []
 synctex = dependency('synctex', version: '>=1.19', required: get_option('synctex'))
 seccomp = dependency('libseccomp', required: get_option('seccomp'))
-#landlock = dependency('linux', version: '>=5.13', required: get_option('landlock'))
+landlock = cc.check_header('linux/landlock.h', required: get_option('landlock'))
 
 if synctex.found()
   build_dependencies += synctex
@@ -95,7 +95,7 @@ if seccomp.found()
   additional_sources += files('zathura/seccomp-filters.c')
 endif
 
-if get_option('landlock').auto()
+if landlock
   defines += '-DWITH_LANDLOCK'
   additional_sources += files('zathura/landlock.c')
 endif
--- a/zathura/landlock.c
+++ b/zathura/landlock.c
@@ -1,7 +1,9 @@
+#if WITH_LANDLOCK
 #define _GNU_SOURCE
 #include <linux/landlock.h>
 #include <sys/prctl.h>
 #include <sys/syscall.h>
+#endif
 #include <stdio.h>
 #include <fcntl.h> 
 #include <unistd.h> 
@@ -71,18 +73,20 @@ landlock_drop (__u64 fs_access)
   LANDLOCK_ACCESS_FS_READ_FILE | \
   LANDLOCK_ACCESS_FS_READ_DIR)
 
+#if WITH_LANDLOCK
 void
 landlock_drop_write (void)
 {
-  landlock_drop(_LANDLOCK_ACCESS_FS_WRITE | \
+  landlock_drop(_LANDLOCK_ACCESS_FS_WRITE |
     LANDLOCK_ACCESS_FS_EXECUTE);
 }
+#endif
 
 void
 landlock_drop_all (void)
 {
-  landlock_drop(_LANDLOCK_ACCESS_FS_READ | \
-    _LANDLOCK_ACCESS_FS_WRITE | \
+  landlock_drop(_LANDLOCK_ACCESS_FS_READ |
+    _LANDLOCK_ACCESS_FS_WRITE |
     LANDLOCK_ACCESS_FS_EXECUTE);
 }
 
--- a/zathura/landlock.c
+++ b/zathura/landlock.c
@@ -1,0 +1,155 @@
+#define _GNU_SOURCE
+#include <linux/landlock.h>
+#include <sys/prctl.h>
+#include <sys/syscall.h>
+#include <stdio.h>
+#include <fcntl.h> // for open
+#include <unistd.h> // for close
+#include <girara/utils.h>
+
+//#include <stdlib.h>
+
+#include "landlock.h"
+
+
+#ifndef landlock_create_ruleset
+static inline int
+landlock_create_ruleset(const struct landlock_ruleset_attr *const attr,
+			const size_t size, const __u32 flags)
+{
+	return syscall(__NR_landlock_create_ruleset, attr, size, flags);
+}
+#endif
+
+#ifndef landlock_add_rule
+static inline int
+landlock_add_rule(const int ruleset_fd,
+		  const enum landlock_rule_type rule_type,
+		  const void *const rule_attr, const __u32 flags)
+{
+	return syscall(__NR_landlock_add_rule, ruleset_fd, rule_type,
+			rule_attr, flags);
+}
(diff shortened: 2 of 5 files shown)


sebastinas:
This looks much more maintainable then the seccomp filters. So could we just drop the seecomp filter and move to landlock? With the ability to restrict on a path-basis, this would also enable the use of dbus and stop crippling the user experience in strict sandbox mode.

valoq:
> This looks much more maintainable then the seccomp filters. So could we just drop the seecomp filter and move to landlock? With the ability to restrict on a path-basis, this would also enable the use of dbus and stop crippling the user experience in strict sandbox mode.

While this certainly provides a simple way to isolate the filesystem, it does not address the kernel attack surface at all. For that we need to filter system calls as well.
That said, it may be a good idea to reintroduce the normal sandbox mode (and make it default again) which could rely only on landlock for filesystem isolation, since that would address most low hanging fruits that an adversary would target.

In order to make a complete sandbox that does not block any features, we would need to split the zathura process and have one sandboxed renderer process that is managed through IPC by a primary zathura process (like openssh and modern browsers do it).
I have been looking into this for a while, but it does require quite a bit of work since splitting the memory management is not easy and requires changes to the plugin API.
The most reasonable approach to this would be something like https://github.com/google/sandboxed-api which could be implemented in libraries like poppler and mupdf. That would also minimized the required changes to zathura and benefit more project that use those libs.

valoq:
For now I kept some of the currently unused code in case we do use it for an alternative sandbox version.

Also credit for most of the code goes to @l0kod and his documentation.

meson.build
```diff
@@ -79,6 +79,7 @@ flags = cc.get_supported_arguments(flags)
 additional_sources = []
 synctex = dependency('synctex', version: '>=1.19', required: get_option('synctex'))
 seccomp = dependency('libseccomp', required: get_option('seccomp'))
+#landlock = dependency('linux', version: '>=5.13', required: get_option('landlock'))
```
sebastinas, line 82: This should check for the availability of the landlock header file.

meson.build
```diff
@@ -94,6 +95,11 @@ if seccomp.found()
   additional_sources += files('zathura/seccomp-filters.c')
 endif
 
+if get_option('landlock').auto()
```
sebastinas, line 98: This needs to be changed based on the availability of the the landlock header.

zathura/landlock.c
```diff
@@ -0,0 +1,129 @@
+#define _GNU_SOURCE
```
sebastinas, line 1: This file is missing the common comments.

zathura/landlock.c
```diff
@@ -0,0 +1,129 @@
...
+#include <linux/landlock.h>
+#include <sys/prctl.h>
+#include <sys/syscall.h>
```
sebastinas, line 4: Should only be included if required to implement the fallbacks.

zathura/landlock.c
```diff
@@ -0,0 +1,129 @@
...
+  LANDLOCK_ACCESS_FS_MAKE_SYM)
+
+#define _LANDLOCK_ACCESS_FS_READ ( \
+  LANDLOCK_ACCESS_FS_READ_FILE | \
+  LANDLOCK_ACCESS_FS_READ_DIR)
+
+void
+landlock_drop_write (void)
+{
+  landlock_drop(_LANDLOCK_ACCESS_FS_WRITE | \
```
sebastinas, line 77: No \ required.

zathura/landlock.c
```diff
@@ -0,0 +1,129 @@
...
+    LANDLOCK_ACCESS_FS_EXECUTE);
+}
+
+void
+landlock_drop_all (void)
+{
+  landlock_drop(_LANDLOCK_ACCESS_FS_READ | \
+    _LANDLOCK_ACCESS_FS_WRITE | \
```
sebastinas, line 85: No \ required.

zathura/landlock.c
```diff
@@ -0,0 +1,129 @@
...
+
+  if (dir_fd == -1)
+    close (path_beneath.parent_fd);
+  close(ruleset_fd);
+}
+
+void
+landlock_restrict_write(void)
+{
+    char* data_path = girara_get_xdg_path(XDG_DATA);
```
sebastinas, line 126: Leaks memory.

zathura/landlock.h
```diff
@@ -0,0 +1,3 @@
+void landlock_drop_write (void);
```
sebastinas, line 1: Missing include guards and common file comments.

sebastinas (changes requested):

sebastinas:
> > This looks much more maintainable then the seccomp filters. So could we just drop the seecomp filter and move to landlock? With the ability to restrict on a path-basis, this would also enable the use of dbus and stop crippling the user experience in strict sandbox mode.
> 
> While this certainly provides a simple way to isolate the filesystem, it does not address the kernel attack surface at all. For that we need to filter system calls as well. That said, it may be a good idea to reintroduce the normal sandbox mode (and make it default again) which could rely only on landlock for filesystem isolation, since that would address most low hanging fruits that an adversary would target.

I am not sure if seccomp is actually a maintainable approach in the long run. The last years have shown that it requires whitelisting of platform specific syscalls and they change without notice depending on the dependencies. It would potentially make more sense to provide flatpak images (or comparable) for some sandboxing without breaking features and a separate zathura-seccomp binary for people that want it.

> In order to make a complete sandbox that does not block any features, we would need to split the zathura process and have one sandboxed renderer process that is managed through IPC by a primary zathura process (like openssh and modern browsers do it). I have been looking into this for a while, but it does require quite a bit of work since splitting the memory management is not easy and requires changes to the plugin API. The most reasonable approach to this would be something like https://github.com/google/sandboxed-api which could be implemented in libraries like poppler and mupdf. That would also minimized the required changes to zathura and benefit more project that use those libs.

The likely hood of that being merged is close to 0.

valoq:
Thanks for the review

> > > This looks much more maintainable then the seccomp filters. So could we just drop the seecomp filter and move to landlock? With the ability to restrict on a path-basis, this would also enable the use of dbus and stop crippling the user experience in strict sandbox mode.
> > 
> > 
> > While this certainly provides a simple way to isolate the filesystem, it does not address the kernel attack surface at all. For that we need to filter system calls as well. That said, it may be a good idea to reintroduce the normal sandbox mode (and make it default again) which could rely only on landlock for filesystem isolation, since that would address most low hanging fruits that an adversary would target.
> 
> I am not sure if seccomp is actually a maintainable approach in the long run. The last years have shown that it requires whitelisting of platform specific syscalls and they change without notice depending on the dependencies.

I see what you mean and frankly I did not expect this feature to require this many adjustments when I first implemented it. Though to em it seems manageable to address upcoming syscall changed introduced by dependencies by addressing new kernel syscalls with the ERRNO_RULE since those are usually the syscalls that are used by newer version of dependent libraries and they always have fallbacks for older kernels included.

> It would potentially make more sense to provide flatpak images (or comparable) for some sandboxing without breaking features and a separate zathura-seccomp binary for people that want it.

A separate binary would address issues such as the default gtk dbus socket, so maybe this is a more reasonable approach. 
As for general sandboxing, there are multiple options available already. Either through mandatory access control like apparmor ([here](https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/profiles-s-z/zathura)) or by using the same isolation technology as flatpak like [this](https://github.com/valoq/bwscripts/blob/master/profiles/zathura).


> > In order to make a complete sandbox that does not block any features, we would need to split the zathura process and have one sandboxed renderer process that is managed through IPC by a primary zathura process (like openssh and modern browsers do it). I have been looking into this for a while, but it does require quite a bit of work since splitting the memory management is not easy and requires changes to the plugin API. The most reasonable approach to this would be something like https://github.com/google/sandboxed-api which could be implemented in libraries like poppler and mupdf. That would also minimized the required changes to zathura and benefit more project that use those libs.
> 
> The likely hood of that being merged is close to 0.

Can you please elaborate on this? Do you mean you don't see a feature like sandboxed-api to make it into projects like poppler or are you opposed to having this implemented as part of zathura? 
As I understand it, with sandboxed-api there would be only small changes required to zathura plugins that want to use that api.
Not to mention it would address the maintability of seccomp since the filter would be applied to the renderer process only that should need very limited and static syscalls.

zathura/landlock.c
valoq: Not sure what you mean by common comments. It is included in the landlock macro now though, so it wont be included in non Linux builds

sebastinas:
```
/* SPDX-License-Identifier: Zlib */
```

zathura/landlock.h
sebastinas: No, this should be
```
#ifndef ZATHURA_LANDLOCK_H
#define ZATHURA_LANDLOCK_H

....

#endif
```

valoq:
Github still shows "Changes Requested" but I believe all issues were addressed. 
@sebastinas if there is anything missing please let me know.

sebastinas:
Haven't had the the time to test the changes yet.

PR #588, opened 2024-06-05
Reset adjust mode when hiding inputbar

9ary:
Zathura sets a special adjust mode before showing the inputbar. Remaining in that mode after it has been hidden breaks things like rotation, so reset to none.


jdujava:
Woudn't it be nice to restore the adjust mode before entering the inputbar?
As of now, the original adjust mode is lost on every search or inputbar action.

9ary:
Yeah, I've thought of this too. This technically isn't a regression since the inputbar was leaving the adjust mode in a broken state before.

I was under the impression that hiding the index did the same thing, but it actually only does it when the adjust mode was inputbar:
https://github.com/pwmt/zathura/blob/b794eb2919ce5b01ffceda941dff65686ea32f73/zathura/shortcuts.c#L1245-L1248

I guess this could be cleaned up, and the adjust mode can be restored.

jdujava:
> Yeah, I've thought of this too. This technically isn't a regression since the inputbar was leaving the adjust mode in a broken state before.

Exactly, this PR seemed to go in the direction of improving this behavior, hence my suggestion.

sebastinas (changes requested):
Please implement the changes mentioned by @jdujava.

9ary:
Sorry, I stopped using Zathura due to growing frustration with some less trivial bugs. I think it'd be best if someone else picked up this patch.

Issue #608, opened 2024-07-08
Conflicting code causes zathura to crop oversized pdf pages


malmerri42:
From issue #602, I forked the development branch and inspected cell_width and cell_height which are part of the document struct defined here:
https://github.com/pwmt/zathura/blob/c3ffe3e02fc65be6de92dfc25a7f77e42c0b4e0e/zathura/document.c#L37-L38

Placing a watchpoint after the document has been created on cell_width and cell_height shows that initially this code:
https://github.com/pwmt/zathura/blob/c3ffe3e02fc65be6de92dfc25a7f77e42c0b4e0e/zathura/adjustment.c#L16-L24
sets the cell_width and cell_height to the dimensions of the largest page in the document. 

Later down the line the following code overwrites the above adjustments to cell_width and cell_height:
https://github.com/pwmt/zathura/blob/c3ffe3e02fc65be6de92dfc25a7f77e42c0b4e0e/zathura/zathura.c#L1114-L1118
It seems like it sets the cell_width and cell_height to the most frequent page width and height.
If you comment out the above code, zathura will correctly show oversized pdf pages.

sebastinas:
Fixed in 9c0017b.

Issue #447, opened 2024-02-06
Fit width mode for variable page sizes


sebastinas:
*On GitLab by @helixoid on Feb 6, 2024, 06:46*

---

I have some books (especially One Piece Manga) which has variable page sizes(several portrait and some landscape/dual-page-spreads) and when I press "s" it adjusts the document according to the width of page in landscape mode, at least it did before not now because of the page cropping issue.

I want it to adjust the document fit on per page basis like how **Okular** does.

I really love this program and not really a fan of all the kde dependencies that come with okular.

I also don't know how make zathura prevent scrolling to the next page when using mouse or touchpad one solution could be to **display one page at a time and one change the page when pressed pagedown key and not the mouse scroll wheel**.

And manga pages(dual-page-spreads) with different page sizes and getting cropped and some of the information is not displayed.

phos-rt:
Today I faced this same problem and tracked it back to this [issue](https://git.pwmt.org/pwmt/zathura/-/issues/116) and corresponding [commit](https://git.pwmt.org/pwmt/zathura/-/commit/da33077766a9adc40d6b31186ed641aefc5f1b5f).

Reverting the changes made in that commit makes Zathura able to render different sized pages correctly again, but reinstates the issues originally reported (inconsistent page padding in some pages and fit commands fitting to largest image in the document).

Not sure what the course of action should be but personally I'd like Zathura to go back on that change and be able to render pages of different sizes without cropping. Maybe we could find other ways to fix the issues in the initial report.
 
Here's a before and after of reverting the change:

zathura 0.5.4
![screen-2024-05-12T22:03:53+0200](https://github.com/pwmt/zathura/assets/164511066/2fe4db6e-797f-47e1-a7ca-a649a82c5172)

zathura 0.5.6 reverting the changes from that commit
![screen-2024-05-12T22:03:29+0200](https://github.com/pwmt/zathura/assets/164511066/156aed68-f97a-4453-a9dc-9d791dcb050e)

devensiv:
related to #436 

Im still having this issue using 0.5.6 (OS Arch Linux)
`zathura --version`
```
zathura 0.5.6
girara 0.4.4 (runtime: 0.4.4)
(plugin) pdf-poppler (0.3.2) (/usr/lib/zathura/libpdf-poppler.so)
```

my current workaround is downgrading to
`zathura --version`
```
zathura 0.5.2
girara 0.3.9 (runtime: 0.3.9)
[...]

PR #609, opened 2024-07-08
deleted code that changed cell_width and cell_height from the max of …

malmerri42:
…all the pages

See #608 

I am new to contributing, please let me know if I made any mistakes.

### DIFF
--- a/zathura/zathura.c
+++ b/zathura/zathura.c
@@ -861,49 +861,6 @@ typedef struct {
   double w;
   int freq;
 } sample_t;
-
-static int document_page_size_comp(const void* a, const void* b) {
-  const sample_t* lhs = a;
-  const sample_t* rhs = b;
-  return rhs->freq - lhs->freq;
-}
-
-static void document_open_page_most_frequent_size(zathura_document_t* document, unsigned int* width,
-                                                  unsigned int* height) {
-  girara_list_t* samples             = girara_list_new2(g_free);
-  const unsigned int number_of_pages = zathura_document_get_number_of_pages(document);
-
-  for (unsigned int page_id = 0; page_id < number_of_pages; ++page_id) {
-    zathura_page_t* page = zathura_document_get_page(document, page_id);
-    const double w       = zathura_page_get_width(page);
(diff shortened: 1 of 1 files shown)


malmerri42:
Note, this make double page view look very spaced out.

zathura/zathura.c
```diff
@@ -1100,21 +1057,12 @@ bool document_open(zathura_t* zathura, const char* path, const char* uri, const
...
     zathura_page_t* page = zathura_document_get_page(document, page_id);
     if (page == NULL) {
       goto error_free;
     }
 
-    unsigned int cell_height = 0, cell_width = 0;
-    zathura_document_get_cell_size(document, &cell_height, &cell_width);
-    zathura_document_set_cell_size(document, most_freq_height, most_freq_width);
-    zathura_page_set_width(page, most_freq_width);
-    zathura_page_set_height(page, most_freq_height);
```
sebastinas, line 1116: The calls to `zathura_page_set_width/height` can probably go away. But the cell size still needs to be set, IMHO. I need to take a closer look at what we wanted to achieve here, but changing the page size here seems wrong in any case.

malmerri42: Hi Sebastian, thanks very much for working on this project.

It seems like the cell size is being initially set here:
https://github.com/pwmt/zathura/blob/c3ffe3e02fc65be6de92dfc25a7f77e42c0b4e0e/zathura/adjustment.c#L16-L24

If I had to guess what the purpose of the code I removed was, is that without it the double page view looks very spaced out because cell_width/height is over the whole document, not page by page. 
So when the cell_width/height is set according to the largest page, that causes the double view to look strange.

phos-rt: fyi this issue was also reported at #447 . On my comment I noted the original issue and commit that added those calls to `zathura_page_set_width()` and `zathura_page_set_height()`, in case they provide you with more insight on why they were added! Thank you for working on a fix ::)

sebastinas:
Fixed differently in 9c0017b.

sebastinas:
The fix was incomplete. I will adapt this PR.

PR #612, opened 2024-07-08
Fix selection/highlighter behavior

### LINKED ISSUE #599: Option for `abort` to clear all highlights
Selecting text (or highlighting when `highlighter-modifier` is pressed) with mouse leaves highlighted regions. Pressing the `Escape` key (thus calling `abort`) clears the search result highlights and also message of what text was copied, but it doesn't clear highlights from selection/highlight actions.

It seems more consistent that the `abort` action would clear all highlights (as it already does for the search ones), or at least it would be nice to have a configuration option for that.

### LINKED ISSUE #611: Weird transition between selection/highlighter
Transitioning between selecting text and highlighting by pressing `highlighter-modifier` mid selection/highlighting doesn't properly clear the previous selected/highlighted region.

Following behavior is also not desired:
- start the selection by dragging with `Button1`
- press `highlighter-modifier` to only highlight a region
- releasing `Button1` **still copies** the text selected before press of `highlighter-modifier`

jdujava:
- clear selected/highlighted region on abort
- proper clearing when transitioning between selection/highlighter

Fixes #599
Fixes #611
Fixes https://github.com/pwmt/zathura-djvu/issues/6

### DIFF
--- a/zathura/page-widget.c
+++ b/zathura/page-widget.c
@@ -990,6 +990,17 @@ static void rotate_point(zathura_document_t* document, double orig_x, double ori
   }
 }
 
+void zathura_page_widget_clear_selection(ZathuraPage* widget) {
+  ZathuraPagePrivate* priv = zathura_page_widget_get_instance_private(widget);
+  if (priv->selection.list != NULL) {
+    girara_list_free(priv->selection.list);
+    priv->selection.list = NULL;
+  }
+  priv->selection.draw   = false;
+  priv->highlighter.draw = false;
+  zathura_page_widget_redraw_canvas(widget);
+}
+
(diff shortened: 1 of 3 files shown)


sebastinas:
This looks good in principle, but I am not quite happy with the different ways to reset to control things now. Somethings are done via functions, some other with properties of the objects achieving the same things.

jdujava:
I am not sure what you mean, since I basically just reshuffled code around so that selection/highlight clearing works properly. Lacking more thorough knowledge about the codebase I dont know which parts would you like me to refactor.

If you give me some pointers, I could try to look at it (maybe in a subsequent PR?).

PR #621, opened 2024-07-09
Implement `jumplist` command

jdujava:
Shows recent jumps in jumplist.

### DIFF
--- a/zathura/commands.c
+++ b/zathura/commands.c
@@ -149,6 +149,56 @@ bool cmd_bookmark_open(girara_session_t* session, girara_list_t* argument_list) 
   return true;
 }
 
+bool cmd_jumplist_list(girara_session_t* session, girara_list_t* argument_list) {
+  g_return_val_if_fail(session != NULL, false);
+  g_return_val_if_fail(session->global.data != NULL, false);
+  zathura_t* zathura = session->global.data;
+  if (zathura->document == NULL) {
+    girara_notify(session, GIRARA_ERROR, _("No document opened."));
+    return false;
+  }
+
+  int num_entries = 5;
+
+  const unsigned int argc = girara_list_size(argument_list);
+  if (argc == 1) {
+    int value = g_ascii_strtoll(girara_list_nth(argument_list, 0), NULL, 10);
+    if (value == 0) {
+      girara_notify(session, GIRARA_WARNING, _("Argument must be a nonzero number."));
+      return false;
+    } else if (value > 0) {
+      num_entries = value;
+    } else {
+      num_entries = (int)zathura->jumplist.size + value;
+    }
+  } else if (argc != 0) {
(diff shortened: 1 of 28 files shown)


zathura/commands.c
```diff
@@ -149,6 +149,69 @@ bool cmd_bookmark_open(girara_session_t* session, girara_list_t* argument_list)
...
+    return false;
+  }
+
+  int num_entries = 5;
+
+  const unsigned int argc = girara_list_size(argument_list);
+  if (argc == 1) {
+    const char* value = girara_list_nth(argument_list, 0);
+    if (value != NULL) {
+      if (atoi(value) == 0) {
```
sebastinas, line 167: The result of this operation can be stored and does not need to be repeated.

Also we mostly use `strto*` to convert strings to integers as it has better error handling.

zathura/commands.c
```diff
@@ -149,6 +149,69 @@ bool cmd_bookmark_open(girara_session_t* session, girara_list_t* argument_list)
...
+  girara_list_iterator_t* iter = girara_list_iterator(zathura->jumplist.list);
+  while (girara_list_iterator_has_next(iter)) {
+    girara_list_iterator_next(iter);
+  }
+
+  GString* string = g_string_new(NULL);
+  for (int i = 0; i < num_entries; ++i) {
+    zathura_jump_t* j = girara_list_iterator_data(iter);
+    g_string_append_printf(string, "[%d]: page=<b>%2d</b>, x=%f, y=%f %s\n", zathura->jumplist.size - i - 1, j->page,
+                           j->x, j->y, girara_list_iterator_data(iter) == current_jump ? "(current)" : "");
```
sebastinas, line 196: page and current should be translatable strings.

zathura/commands.c
```diff
@@ -149,6 +149,69 @@ bool cmd_bookmark_open(girara_session_t* session, girara_list_t* argument_list)
...
+
+  if (num_entries <= 0) {
+    return true;
+  }
+
+  zathura_jump_t* current_jump = zathura_jumplist_current(zathura);
+  girara_list_iterator_t* iter = girara_list_iterator(zathura->jumplist.list);
+  while (girara_list_iterator_has_next(iter)) {
+    girara_list_iterator_next(iter);
+  }
```
sebastinas, line 190: This loop and the next loop can be simplified by iterating from `size - 1` down.  The use of iterators should mostly go away.

sebastinas (changes requested):

jdujava:
Tried to implement all of the suggestions. Glad to hear any additional feedback (mainly just trying to adapt already existing code).

zathura/commands.c
jdujava: Not sure if I regenerated all necessary files correctly. If not, pointers to documentation or necesarry commands would be helpful.

PR #628, opened 2024-07-11
Refactor/fix current index row

### LINKED ISSUE #613: Wrong current index row
The current index row selection implemented in #561 is flawed. Consider following scenario:

**Index**
```
1  First Section
  1.1  Subsection One
  1.2  Subsection Two
2  Second Section
```

Scrolling the document right before `2  Second Section`, the "current index row" is then chosen as `1  First Section`.

jdujava:
Refactor and simplify by remembering the last index path with `index path page number` <= `current page`.

Fix #613.

### DIFF
--- a/zathura/utils.c
+++ b/zathura/utils.c
@@ -108,42 +108,36 @@ void document_index_build(girara_session_t* session, GtkTreeModel* model, GtkTre
   }
 }
 
-gboolean _for_each_func(GtkTreeModel* model, GtkTreePath* path, GtkTreeIter* iter, gpointer data) {
-  zathura_t* zathura = (zathura_t*)data;
+typedef struct {
+  GtkTreeIter current_iter;
+  unsigned int current_page;
+} SearchData;
+
+static gboolean search_current_index(GtkTreeModel* model, GtkTreePath* UNUSED(path), GtkTreeIter* iter, gpointer data) {
+  SearchData* search_data = (SearchData*)data;
   zathura_index_element_t* index_element;
   gtk_tree_model_get(model, iter, 3, &index_element, -1);
   zathura_link_target_t target = zathura_link_get_target(index_element->link);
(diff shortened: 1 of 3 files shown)


zathura/utils.c
```diff
@@ -108,42 +108,27 @@ void document_index_build(girara_session_t* session, GtkTreeModel* model, GtkTre
   }
 }
 
-gboolean _for_each_func(GtkTreeModel* model, GtkTreePath* path, GtkTreeIter* iter, gpointer data) {
+gboolean _update_current_index_path(GtkTreeModel* model, GtkTreePath* path, GtkTreeIter* iter, gpointer data) {
```
sebastinas, line 111: This function could be static.

zathura/utils.c
```diff
@@ -108,42 +108,27 @@ void document_index_build(girara_session_t* session, GtkTreeModel* model, GtkTre
...
-      path = gtk_tree_model_get_path(model, &parent_iter);
-    }
-    gtk_tree_view_expand_to_path(tree_view, path);
-    gtk_tree_view_set_cursor(tree_view, path, NULL, FALSE);
-    gtk_tree_view_scroll_to_cell(tree_view, path, NULL, TRUE, 0.5, 0.0);
-
-    return TRUE;
+    return TRUE; // terminate foreach loop
   }
+  zathura->global.current_index_path = gtk_tree_path_copy(path);
```
sebastinas, line 120: Does this need to be freed at some point?

jdujava: Thanks, hopefully now it is OK.

sebastinas: I would also expect a `gtk_tree_path_free` here if `zathura->global.current_index_path` is not `NULL`.

jdujava: Yeah, of course..

jdujava:
Pushed even better (in my opinion) refactor.

PR #638, opened 2024-07-17
Add a template for bug report and feature request in Github repo

laomuon:
This PR is to add a template when creating a bug report or a feature request to the Github repo, so that the one who create it can provide all the informations needed

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,1 +1,1 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false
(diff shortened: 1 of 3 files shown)


laomuon:
@sebastinas I think this can be a positive thing to the development of the repo without any modification to the codebase

sebastinas:
Can this be implemented on a organization level instead?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
alerque, line 1: Please don't do this. I understand why templates help most people remember what they should include in a good bug report, but disabling this means people like me can't use extensive history as FOSS maintainers to know what we should write and use CLI tools to manage bugs.

laomuon:
> Can this be implemented on a organization level instead?

I think it will be better if every repo has its own template

sebastinas:
> > Can this be implemented on a organization level instead?
> 
> I think it will be better if every repo has its own template

Why though? It's essentially the same for all the repos.

sebastinas:
This has been implemented on the organization level.

PR #641, opened 2024-07-21
fix: hide page label if it is equal to page number

### LINKED ISSUE #590: An option to switch `[current_page / total_pages]` to just `total_pages`
zathura-poppler (and now zathura-mupdf) seemingly provide total pages count as (current_page/total_pages) so the statusbar usually displays something like `[33 / (33/273)]`.

I'm using Arch Linux with the following versions of software:
```
info: Opening plain database via sqlite backend.
warning: sqlite database already exists. Set your database backend to sqlite
zathura 0.5.6
girara 0.4.4 (runtime: 0.4.4)
(plugin) djvu (0.2.9) (/usr/lib/zathura/libdjvu.so)
(plugin) pdf-mupdf (0.4.2) (/usr/lib/zathura/libpdf-mupdf.so)
```

VishnuSanal:
fixes #590

### DIFF
--- a/zathura/zathura.c
+++ b/zathura/zathura.c
@@ -1674,7 +1674,12 @@ statusbar_page_number_update(zathura_t* zathura)
       if (show_percent) {
         page_number_text = g_strdup_printf("[%s (%d/%d) (%d%%)]", page_label, current_page_number + 1, number_of_pages, page_number_percent);
       } else {
-        page_number_text = g_strdup_printf("[%s (%d/%d)]", page_label, current_page_number + 1, number_of_pages);
+        char page_number_string[5];
+        sprintf(page_number_string, "%d", current_page_number + 1);
+        if (strcmp(page_label, page_number_string))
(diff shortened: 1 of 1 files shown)


sebastinas (changes requested):
It would be cleaner (and would also cover all cases) if in `zathura_page_new` `label` is not set if the label matches the page number. Also, please make no assumptions here on how `current_page_number` may or may not become and use `g_ascii_dtostr` instead - you can find uses of it in the codebase.

VishnuSanal:
hi Sebastian, thanks for you review! but, I didn't quite get what you meant. can you please clarify the following?

> It would be cleaner (and would also cover all cases) if in `zathura_page_new` `label` is not set if the label matches the page number.

- did you mean to revert my current implementation & to hide the label completely from `zathura_page_new` function in `page.c` if `page_label` equals the `page_number`?

> Also, please make no assumptions here on how `current_page_number` may or may not become and use `g_ascii_dtostr` instead - you can find uses of it in the codebase.

- sure, will do.

Issue #603, opened 2024-06-18
Position lost when reloading file


ghbrown:
Like many other users, I use Zathura as my PDF viewer for LaTeX writing, so Zathura has the PDF open when I rebuild the PDF.

Since my latest update via `pacman` Zathura no longer holds the proper position in the file when the PDF is rebuilt.

However, if I open a file, navigate to position X, close Zathura, then reopen the file it opens to X no problem. Seems there's something special about the rebuild of an open file that's causing issues for position.

System info:
```
EndeavourOS
zathura 0.5.6
girara 0.4.4 (runtime: 0.4.4)
(plugin) pdf-poppler (0.3.2) (/usr/lib/zathura/libpdf-poppler.so)
```

sebastinas:
Please provide a log (with `-l debug`) with a4ef953 from the develop branch or newer.

ghbrown:
Working on this now. How do I find the correct path for a plugin?
I'm currently running the newly compiled version like
```bash
<absolute path>/build/zathura -p /usr/lib/zathura main.pdf
```
I'm using this plugin path since it's what my installed zathura returns from `zathura --version`, but it doesn't seem to be working:
```
warning: Found no plugins. Please install at least one plugin.
info: Opening plain database via sqlite backend.
warning: sqlite database already exists. Set your database backend to sqlite. See zathura
rc(5) for details.
error: Could not determine file type.
```

sebastinas:
Right, the path is correct, but the plugin ABI changed. You'll need to rebuild the plugins against the zathura version from `develop` (best to also use the develop version of the plugins).

ghbrown:
When you say "against the zathura version" from the develop branch, what do you mean?
I've build `zathura-pdf-poppler` according to its build instructions, and I'm now passing this path (which contains `libpdf-poppler.so`), but I'm getting the same error.

I think perhaps the answer to both is that I haven't `ninja install`ed zathura (just to keep my machine as clean as possible). Is this what is needed before building `zathura-pdf-poppler`?

ghbrown:
Indeed that was what I needed to do. Attached is the log.

I opened the PDF, scrolled down a bit, rebuilt via `latexmk` (which generates a new pdf). The bug was present, so it went back to the beginning.
[debug.log](https://github.com/user-attachments/files/15946353/debug.log)

sebastinas:
[...]

PR #642, opened 2024-07-22
fix: fix lost file info when reloading the file after name changing

laomuon:
Fix lost file info when reloading the file after name changing by try to reload the info from `predecessor_document`. 
Issue: #603


sebastinas (changes requested):
Looks good, but please remove the whitespace change at the end of `zathura/zathura.c` from the PR.

laomuon:
@sebastinas Done!

PR #647, opened 2024-07-29
Fix flickering page number

### LINKED ISSUE #646: Flickering page number when scrolling with dual-page layout
When scrolling with dual-page layout (positioned in the exact center, for example obtained by best-fit adjustment), the page number in statusbar is flickering between the values of the two visible pages. Also every `j`/`k` switches between the pages in the statusbar (again, when positioned in the exact center).

jdujava:
Previously, when scrolling with dual-page layout positioned in the exact center (for example obtained by best-fit adjustment), the boundary handling could change horizontal alignment in such a way that resulted in flickering of the page number between two visible pages.

This simple fix seems to work, however, I am not sure whether this is the best way. Other suggestions are welcome.

Fixes #646.


sebastinas (changes requested):
Could you please add a comment why this is added to the code? Thanks

sebastinas (approved):
Thanks @jdujava

Issue #450, opened 2024-02-27
[Question] How to display search match counters?


sebastinas:
*On GitLab by @itaranto on Feb 27, 2024, 16:53*

---

Hello, 

Is there a way to display search match counters? 

I'd like to see something like [1/5], which would mean the current match and the total matches.

I read the documentation and I couldn't find this functionality, so I'm not sure if it exists or not.

sebastinas:
That is currently not implemented.

appryst:
Hey, is that feature still on the agenda? :)

appryst:
Not wanting to sound negative, but is that realistic given that the PR is over 1.5 years old without change?

PR #648, opened 2024-07-29
search once

### LINKED ISSUE #132: Crazy memory usage
*On GitLab by @alyssoncs on Jun 14, 2019, 22:34*

---

When searching through a pdf document the zathura's memory usage goes crazy.

Searching the word "service" in [Android Notes for Professionals](https://books.goalkicker.com/AndroidBook/) for example, makes zathura use 4GiB of memory in my machine.

OS: Debian 10

zathura version: 0.4.3

zathura-pdf-poppler version: 0.2.9

ldwgchen:
Instead of searching the whole document in one go, search and stop at the nearest page that produces a hit.

Fixes #132. UX won't be much different, besides that when the user views a page that hasn't been searched before, they will no longer see highlighted results in this page.

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -76,7 +76,8 @@ void update_visible_pages(zathura_t* zathura) {
       GObject* obj_page_widget = G_OBJECT(page_widget);
       g_object_get(obj_page_widget, "search-results", &results, NULL);
       if (results != NULL) {
-        g_object_set(obj_page_widget, "search-current", 0, NULL);
+        g_object_set(obj_page_widget, "search-current",
+                     zathura->global.search_direction == FORWARD ? 0 : girara_list_size(results) - 1, NULL);
       }
     }
   }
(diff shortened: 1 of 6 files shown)


ldwgchen:
Possible conflict with #450 since not all results are obtained at once.

alerque:
I would consider this a UX regression. I regularly use the search as a way to view the total instances of a key word across book length texts and see the dispersion.

image

If search results were *only* incremental this would be much clumsier to achieve. Returning the first result and letting the UI unlock would be nice, but the remaining instances should probably be found and highlighted in the background (and an "N of X" counter in the status bar wouldn't be a bad idea either.

laomuon:
With the Vim-like command (`/`), I expect more or less it works like in Vim, i.e. search the whole document and have a counter for it. I agree with @alerque that this could be an UX regression.

ldwgchen:
I'm personally using zathura-pdf-mupdf. I tried opening a problematic document with mupdf and found the resulting memory usage to be quite similar to what I'd get with zathura-pdf-mupdf. Thus, I believe that the root cause of #132 likely lies upstream. Since search-once is disruptive to the common use cases mentioned by @alerque and @laomuon, it's best to look for an alternative solution and close this PR for now.

PR #649, opened 2024-07-30
Preserve search state on reload

### LINKED ISSUE #598: Search string not preserved on reload
Similarly as mentioned in https://github.com/pwmt/zathura/issues/308#issuecomment-2086890721, I use Zathura with LaTeX to continuously reload the pdf file. However, on reload the current search string is not preserved, which can potentially disrupt the workflow.

It would be really nice to preserve the current search string on reload (at least for the current Zathura instance, no need to store it in a database in my opinion).

fcolona:
With this, when reloading the document, the search state is preserved, including the search string and the search highlight. 
As I am new to contributing, review and suggestions are appreciated.

Closes #598

### DIFF
--- a/zathura/commands.c
+++ b/zathura/commands.c
@@ -401,7 +401,8 @@ bool cmd_search(girara_session_t* session, const char* input, girara_argument_t*
   g_return_val_if_fail(session->global.data != NULL, false);
   zathura_t* zathura = session->global.data;
 
-  if (zathura->document == NULL || strlen(input) == 0) {
+  size_t input_len = strlen(input);
+  if (zathura->document == NULL || input_len == 0) {
     return false;
   }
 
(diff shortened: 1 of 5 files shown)


sebastinas:
Please rebase on top of develop.

fcolona:
Hi @sebastinas,
I forgot to tell you that I did the rebase and the PR was ready for review. Could you take a look at it when you have some time? Thank you!

unrelated-stardust:
What is left to do to merge this pull request? I'd like to help.

zathura/commands.c
```diff
@@ -401,15 +401,38 @@ bool cmd_search(girara_session_t* session, const char* input, girara_argument_t*
...
+  }
+  else if(input_len > search_string_len){
+    zathura->global.search_string = g_try_realloc(zathura->global.search_string, search_string_len + input_len + 1);
+  } 
+  /* check whether the memory allocation was successful */
+  if(zathura->global.search_string == NULL){
+    return false;
+  }
+  /* set search string */
+  strcpy(zathura->global.search_string, input);
```
sebastinas, line 434: This is missing the length of the string.

zathura/shortcuts.c
```diff
@@ -76,6 +76,61 @@ static bool link_shortcuts(zathura_t* zathura, girara_callback_inputbar_activate
   return false;
 }
 
+/* Helper function to redo the search and stay in place */
+bool redo_search(zathura_t* zathura){
+    if (zathura->document == NULL || strlen(zathura->global.search_string) == 0) {
```
sebastinas, line 81: First, this function should be static. Second, it is never called with document == NULL - otherwise the call site is buggy. The string length check could also be done at the call site where it is checked if search_string != NULL.

sebastinas (changes requested):

fcolona:
Hi @unrelated-stardust, thanks for picking this up. And thanks to @sebastinas for the review. You are right about the `strcpy` issue, but I was looking at the dynamic memory allocation code for the `zathura->global.search_string` and I thought that maybe we could replace the whole if/else allocation/reallocation logic and the `strcpy` with the `g_strdup` function. What do you think? It would be something like this:

```
  g_free(zathura->global.search_string);
  zathura->global.search_string = g_strdup(input);
  if(zathura->global.search_string == NULL){
    return false;
  }
```

sebastinas:
Sorry for taking so long to respond. I have rebased your PR and adopted it to the changes on develop.

PR #659, opened 2024-08-06
Fix centering current index

### LINKED ISSUE #615: Centering of current index row is wrong on the first index toggle
Toggling index chooses index row corresponding to the current page, and centers it in the index view.

However, on the first opening of the index, the current index row is top aligned (it is necesarry to have long enough index to notice such behavior). Afterwards, the centering works fine.

I have no idea why this happens, any pointers would be welcome.

laomuon:
Fixes: #615 by making the scrolling to be execute after everything else

### DIFF
--- a/zathura/utils.c
+++ b/zathura/utils.c
@@ -141,6 +141,7 @@ void index_scroll_to_current_page(zathura_t* zathura) {
   gtk_tree_model_foreach(model, search_current_index, &search_data);
 
   GtkTreePath* current_path = gtk_tree_model_get_path(model, &search_data.current_iter);
+  g_free(zathura->global.current_index_path);
   zathura->global.current_index_path = gtk_tree_path_to_string(current_path);
   gtk_tree_view_expand_to_path(tree_view, current_path);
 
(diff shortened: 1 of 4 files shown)


jdujava:
Seems to work alright, thanks for looking into this!

zathura/utils.c
```diff
@@ -134,9 +141,11 @@ void index_scroll_to_current_page(zathura_t* zathura) {
   gtk_tree_model_foreach(model, search_current_index, &search_data);
 
   GtkTreePath* current_path = gtk_tree_model_get_path(model, &search_data.current_iter);
+  zathura->global.current_index_path = gtk_tree_path_to_string(current_path);
```
sebastinas, line 144: Does this need to be freed at some point?

laomuon: I assume here that you want for it to be freed each time a new value is assigned

sebastinas: Yes, and when cleaning up.

zathura/zathura.h
```diff
@@ -3,6 +3,7 @@
 #ifndef ZATHURA_H
 #define ZATHURA_H
 
+#include "glib.h"
```
sebastinas, line 6: This is not required here.

PR #677, opened 2024-09-07
First step at enabling searching in index page

### LINKED ISSUE #565: Searching in index
It would be awesome to have search functionality similar to `/` available in the index.

By accident I sometimes make a "doubleclick" `<Mouse3>` in the index, after which typing seems to trigger some kind of "search" functionality. Seems kinda strange, since it isn't documented, coudn't find any further information, and also works only if the given index entry starts with the given search prompt.

Seems to me like a useful thing to have, maybe even some kind of fuzzy matching would be sometimes desirable.

laomuon:
This PR is a proof of concept for enabling search in index page. It only use the native functionality of `GTK`.

The behavior:
- Open the index page, press `C-f` (the default shortcut of `GTK`) to have the interactive search window appear
- Start typing the keyword, the index's cursor will jump to the first occurence,
- Disable interactive search by default so that when the search window is not present, all the keywords works normally (As opposed to before)

Partly fixes #565

### DIFF
--- a/zathura/shortcuts.c
+++ b/zathura/shortcuts.c
@@ -1203,6 +1203,8 @@ bool sc_toggle_index(girara_session_t* session, girara_argument_t* UNUSED(argume
     g_object_set(G_OBJECT(renderer), "ellipsize", PANGO_ELLIPSIZE_END, NULL);
     g_object_set(G_OBJECT(gtk_tree_view_get_column(GTK_TREE_VIEW(treeview), 0)), "expand", TRUE, NULL);
     gtk_tree_view_column_set_alignment(gtk_tree_view_get_column(GTK_TREE_VIEW(treeview), 1), 1.0f);
+    gtk_tree_view_set_search_equal_func(GTK_TREE_VIEW(treeview), search_equal_func_index, treeview, NULL);
+    gtk_tree_view_set_enable_search(GTK_TREE_VIEW(treeview), FALSE);
     g_signal_connect(G_OBJECT(treeview), "row-activated", G_CALLBACK(cb_index_row_activated), zathura);
 
     gtk_container_add(GTK_CONTAINER(zathura->ui.index), treeview);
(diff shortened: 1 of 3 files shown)


zathura/utils.c
```diff
@@ -125,6 +128,82 @@ static gboolean search_current_index(GtkTreeModel* model, GtkTreePath* UNUSED(pa
...
+    return TRUE;
+  }
+  if (find_substring(source_string, key)) {
+    return FALSE;
+  }
+  return TRUE;
+}
+
+gboolean search_equal_func_index(GtkTreeModel* model, gint column, const gchar* key, GtkTreeIter* iter,
+                                 gpointer search_data) {
```
laomuon, line 175: Reimplement the search equal function so that it will also compared with collapsed entries

zathura/utils.c
```diff
@@ -125,6 +128,82 @@ static gboolean search_current_index(GtkTreeModel* model, GtkTreePath* UNUSED(pa
   return FALSE;
 }
 
+static bool find_substring(const char* source_str, const char* search_str) {
```
laomuon, line 131: Find the `search_str` in `source_str`, as opposed to only match from the beginning of `source_str`

zathura/shortcuts.c
```diff
@@ -1203,6 +1203,8 @@ bool sc_toggle_index(girara_session_t* session, girara_argument_t* UNUSED(argume
     g_object_set(G_OBJECT(renderer), "ellipsize", PANGO_ELLIPSIZE_END, NULL);
     g_object_set(G_OBJECT(gtk_tree_view_get_column(GTK_TREE_VIEW(treeview), 0)), "expand", TRUE, NULL);
     gtk_tree_view_column_set_alignment(gtk_tree_view_get_column(GTK_TREE_VIEW(treeview), 1), 1.0f);
+    gtk_tree_view_set_search_equal_func(GTK_TREE_VIEW(treeview), search_equal_func_index, treeview, NULL);
+    gtk_tree_view_set_enable_search(GTK_TREE_VIEW(treeview), FALSE);
```
laomuon, line 1207: Disable interactive search by default, so that all the hotkeys (j, k, q...) are not consummed by the search function

MightyPlaza:
not sure why but after this commit, I cannot use the arrow keys in place of the vim keys to expand and collapse entries

PR #689, opened 2024-10-10
Faster SyncTeX by reusing scanner object

### LINKED ISSUE #685: Speeding up repeated synctex queries?
Currently, zathura reloads the synctex file for every query.

As can be seen in `zathura/synctex.c`:

```cpp
bool synctex_get_input_line_column(const char* filename, unsigned int page, int x, int y, char** input_file,
                                   unsigned int* line, unsigned int* column) {
  if (filename == NULL) {
    return false;
  }

  synctex_scanner_p scanner = synctex_scanner_new_with_output_file(filename, NULL, 1);

...

  synctex_scanner_free(scanner);

  return ret;
}
```

This is very slow for large PDF and synctex files e.g. files with 1000+ lines.

I think it would be preferable if zathura keep the scanner around.

This has the additional complexity of having to watch the synctex file for changes / deletion, but I think this is not a big issue.

If you add

```cpp
#include <synctex/synctex_parser_utils.h>
```

then you can use the variables

```cpp
extern const char * synctex_suffix;
extern const char * synctex_suffix_gz;
```

and function

```cpp
void _synctex_strip_last_path_extension(char * string);
```

in order to compute the file name of the synctex file.

That said, we still need to use either the file monitor or `hash_file_sha256` to detect the changes in the file. The latter is probably easier because it avoids possible race condition with multiple threads accessing the synctex scanner at the same time.

user202729:
Fixes #685 .

The code is mostly copied from synctex's own code for new interactive feature.

A disadvantage I can think of is the scanner object consumes memory even when user is not looking up anything, but I think this is inevitable.

### DIFF
--- a/zathura/synctex.c
+++ b/zathura/synctex.c
@@ -17,85 +17,82 @@
 #include "adjustment.h"
 
 #ifdef WITH_SYNCTEX
-synctex_scanner_p scanner = NULL; // the last scanner object
-time_t last_modification_time = 0; // the last modification time of the synctex file
-char* last_pdf_file_name = NULL; // the last output file name
 
 // Get the modification time of the synctex file of the current scanner.
-time_t get_synctex_file_modification_time() {
+time_t get_synctex_file_modification_time(zathura_t *zathura) {
   struct stat attr;
-  if (stat(synctex_scanner_get_synctex(scanner), &attr) < 0) {
+  if (stat(synctex_scanner_get_synctex(zathura->synctex.scanner), &attr) < 0) {
     return 0;
   }
   return attr.st_mtime;
 }
 
-bool synctex_need_new_scanner(const char* filename) {
-  if (scanner == NULL || last_pdf_file_name == NULL) {
+bool synctex_need_new_scanner(zathura_t *zathura, const char* pdf_filename) {
+  if (zathura->synctex.scanner == NULL || zathura->synctex.last_pdf_file_name == NULL) {
     return true;
   }
 
-  if (g_strcmp0(last_pdf_file_name, filename) != 0) {
+  if (g_strcmp0(zathura->synctex.last_pdf_file_name, pdf_filename) != 0) {
     return true;
   }
 
-  if (get_synctex_file_modification_time() != last_modification_time) {
+  if (get_synctex_file_modification_time(zathura) != zathura->synctex.last_modification_time) {
     return true;
   }
 
   return false;
 }
 
-// Create scanner global variable from given PDF file name.
-bool synctex_make_scanner(const char* filename) {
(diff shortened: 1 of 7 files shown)


zathura/synctex.c
```diff
@@ -16,22 +17,74 @@
 #include "adjustment.h"
 
 #ifdef WITH_SYNCTEX
+synctex_scanner_p scanner = NULL; // the last scanner object
+time_t last_modification_time = 0; // the last modification time of the synctex file
+char* last_pdf_file_name = NULL; // the last output file name
```
sebastinas, line 22: This state would need to go to the `zathura_t` struct. When the document is closed and reopened, also the scanners can be reopened.

sebastinas (changes requested):

zathura/synctex.c
user202729: What do you mean by "the scanners can be reopened"?

I don't think reloading the synctex file early is a good idea consider that for a large document it can take up to a few seconds, which would make zathura look sluggish.

An alternative is to load it in a separate thread — that way CPU is consumed but otherwise nothing bad happens. What do you think?

zathura/zathura.h
```diff
@@ -223,6 +227,15 @@ struct zathura_s {
    * Context for MIME type detection
    */
   zathura_content_type_context_t* content_type_context;
+
+  /**
+   * SyncTeX context. The scanner object is cached for better performance.
+   */
+  struct {
+    synctex_scanner_p scanner;
+    time_t last_modification_time; // the last modification time of the synctex file
```
sebastinas: This doesn't need to be tracked. The idea would be that if the file is reloaded due to changes, the scanner is freed. The next time the scanner is used, it can be reopened.

user202729: Makes sense.

Do you think it's worth considering the possibility that (somehow) only the synctex file is updated and not the PDF (e.g. the user surgically modify the synctex file somehow)?

What I think: in practice it will never happen, and the most common cases (e.g. the user delete the synctex file, a new synctex file appear because user move it from another directory) are gracefully handled.

user202729: Alright, done.

Remark: `sc_reload` is called on PDF file change and `zathura_free` is called on program exit, but both of them calls `document_close` anyway.

Side note 2, technically the non-null check is redundant because `synctex_scanner_free` does the check anyway.

zathura/synctex.c
sebastinas: I meant that when the document is closed, the scanner is closed as well. When the file is reopened and on the next synctex operation, the scanner is then also reloaded.

user202729:
@sebastinas Thanks!

PR #717, opened 2025-01-18
document.c: exit when file is not found

aryalaadi:
It is annoying to deal with 10 windows opening when I accidentally copy a file name that has spaces without quotes.


sebastinas:
So that means that if people mistype an `:open` the program closes. This is the wrong solution for your problem.

aryalaadi:
I still want to fix this, would you merge if I add this check while parsing the command line in main.c instead?

sebastinas:
How though? zathura accepts more than local filenames which make it impossible to check for all cases.

aryalaadi:
guess ill keep this patch for myself

Issue #719, opened 2025-01-19
Handling Long Documents


jchilds0:
Currently loading large documents can cause issues due to overflows in cairo, reported in #449, #365, #253, #200 and #65. Instead of using GtkGrid, either

1.  Create subsurfaces (for each page) of a single surface using `cairo_surface_create_for_rectangle` and draw as many pages as can fit to this surface. Then move the pages around, changing which ones are drawn and where, on this surface as the user scrolls, or
2. Limit the number of pages in the grid, changing which pages are contained in the grid as the user scrolls,

This issue is for any comments since this is not a small change to make, which one seems like the better option and would this resolve those issues.

Issue #449, opened 2024-02-27
Zooming in on long pdfs causes memory leak


sebastinas:
*On GitLab by @sungodmoth on Feb 27, 2024, 16:04*

---

(Note: This is similar to the issue reported in the (very old) issue #209. However, the reporter of that issue did not seem to encounter the boundless increase in memory usage that I have seen, and did not identify the problem as being restricted to long pdfs.) 

Environment: Running i3wm on Manjaro Linux. zathura version 0.5.4, `pdf-mupdf` version 0.4.1.

Description: With most pdfs I am able to zoom as far as zathura allows me to, with no issue. But on _long_ pdfs, seemingly those with \>1000 or so pages, zooming far enough in causes a black screen, and zooming even further than that causes the image to return (but blurry) and zathura's memory usage, as reported by system utilities, _rapidly_ rises (I'm talking about a rate of over a gigabyte per second) with no sign of stopping, until either zathura is closed or the document is again zoomed out, at which point it returns to normal.

Note that this is completely uncorrelated to the file size of the file, and seems to be solely caused by the _page count_. Zooming in on a pdf around 150MB in size but only 500 pages in length, the problem is absent. I have attached for reference a tiny 2000 page file (created by repeated self-concatenation of a minimal pdf) which is only 262KB in size, which displays the problem clearly. 

[long.pdf](https://git.pwmt.org/pwmt/zathura/uploads/1e6fb085bbbb81d1f28f3fb2d5e3c9e3/long.pdf)

(Note that this file throws some warnings, but they are irrelevant to this issue, which I experience with all pdfs of this length). The maximum amount that I can zoom before the black screen appears seems to decrease with the length of the pdf. If testing this is necessary, I've also attached a pdf of twice the length (but still tiny file size).

[longer.pdf](https://git.pwmt.org/pwmt/zathura/uploads/89d7bdd58170c70bbadbc87c589aa2cb/longer.pdf)

jchilds0:
From what I can see the pages are all queued when the screen goes blank, which is caused by the current page not making it to `render_job`. If you keep zooming in, all of the pages begin rendering, which causes the excessive memory usage. 

There may be something overflowing when calculating which pages to render when there is a large page count.

jonathan-sung:
[...]

Issue #365, opened 2022-12-15
Large PDF (ARMv8 reference manual) renders black pages with mupdf backend


sebastinas:
*On GitLab by @ddevault on Dec 15, 2022, 19:42*

---

You may acquire this PDF (12,000 pages) here:

https://developer.arm.com/documentation/ddi0487/latest

sebastinas:
*On GitLab by @valoq on Jan 1, 2023, 01:56*

---

Also reproducible with the poppler backend.

Opening the file in evince with the same poppler version works fine


While zathura just shows a black window when opening the file with zathura-pdf-poppler I could sometimes get the file to show by changing the window size. But even then the application was very slow in response to navigation.

When opening the file with zathura-pdf-mupdf, I also got the following errors multiple times:

```
warning: cannot load object (152728 0 R) into cache
error: calloc (1 x 80 bytes) failed
...
warning: cannot load object (152757 0 R) into cache
error: calloc (16 x 8 bytes) failed
warning: cannot load object (152757 0 R) into cache
error: malloc of 128 bytes failed
...
warning: cannot load object (33456 0 R) into cache
error: calloc (4104 x 1 bytes) failed
error: aborting process from uncaught error!
```

Note that the file opens correctly in Mupdf as well

sebastinas:
*On GitLab by @andipersti on Apr 29, 2023, 18:11*

---

I think I've tracked down the problem to a limitation of the cairo library GTK uses for drawing the widgets.

Zathura basically concatenates all pages of a document into a single big image (in case of the ARM manual the resulting image is about 12.000.000 pixels high in the default zoom level) and uses a sliding window for showing the current view. Now during the GTK drawing process in [`gtk_widget_draw_internal`](https://gitlab.gnome.org/GNOME/gtk/-/blob/3.24.37/gtk/gtkwidget.c?ref_type=tags#L7021-7039) that "clip box" is calculated with `cairo_rectangle` which uses [`_cairo_fixed_from_double`](https://cgit.freedesktop.org/cairo/tree/src/cairo-fixed-private.h?h=1.17.8#n115) for converting the coordinates into the internal fixed-point representation. And this conversion overflows for values bigger than 8.300.000. So the "clip box" will be empty and as a consequence `gdk_cairo_get_clip_rectangle` will return nothing (i.e. an empty box) and the widget won't be drawn.

[...]

Issue #253, opened 2021-02-02
Lots of memory usage when zoomed in


sebastinas:
*On GitLab by @oliverlew on Feb 2, 2021, 03:35*

---

When I open a document the memory usage is normally ~100MB. But if I zoom in to the maximum level, the memory usage goes above 2GB every time. It drops back to ~100MB if I zoom out to a normal level. So I guess that much memory are used to render an extremely zoomed in pdf page. I can reproduce this with almost every pdf document.
zathura version is 0.4.7

PR #721, opened 2025-01-25
ui: dynamic page allocation to gtk grid

jchilds0:
Implements #719 to resolve in #449, #365, #253, #200, and #65

Tested on the ARM reference manual (15,000 pages) https://developer.arm.com/documentation/ddi0487/latest, and Intel manual (5000 pages) https://developer.arm.com/documentation/ddi0487/latest, and doesn't crash when zooming in to either document. 

I'm not sure if the document widget should stay as a GTK widgt or change to a normal struct, I'm fine with either. 

Care needs to be taken when manually setting the value of the ui viewport, since the document widget changes the grid. `zathura_document_widget_[set_ratio_from_value|get_ratio]` functions wrap `zathura_adjustment_[set_ratio_from_value|get_ratio]` functions to handle the conversion from document ratio to widget ratio.

The canvas size chosen is 32,767 which is the max cairo canvas height (perhaps this can be increased but I don't know how much cairo can handle). `zathura_document_widget_render` updates the pages in the grid when the current page is within 1 page of the current pages shown. There's probably a smarter way to handle this as it doesn't do to well with scrolling in zoomed out documents.

### DIFF
--- a/zathura/document-widget.c
+++ b/zathura/document-widget.c
@@ -1,82 +1,49 @@
 /* SPDX-License-Identifier: Zlib */
 
-#include "cairo.h"
-#include "gdk/gdk.h"
+#include <girara/log.h>
+
 #include "glib.h"
 #include "glibconfig.h"
 #include "gtk/gtk.h"
 #include "glib-object.h"
-#include "zathura/document.h"
-#include "zathura/page-widget.h"
-#include "zathura/page.h"
-#include "zathura/types.h"
-#include "zathura/zathura.h"
-#include <girara/log.h>
+#include "adjustment.h"
+#include "document.h"
 #include "document-widget.h"
+#include "page.h"
+#include "types.h"
+#include "zathura.h"
+
+int64_t cairo_max_size = INT16_MAX;
 
 typedef struct zathura_document_widget_private_s {
-  zathura_t *zathura;
   unsigned int spacing;
   unsigned int pages_per_row;
   unsigned int first_page_column;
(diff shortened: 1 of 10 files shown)


jchilds0:
Commit https://github.com/pwmt/zathura/pull/721/commits/16c54989eeb44ddaa76e302496a031686d919c12 updates `zathura_dcument_widget_render` to find the range of visible pages using `page_is_visible` and update the document widget when the range of visible pages is outside the current document widget. This resolves scrolling while zoomed out.

zathura/document-widget.c
```diff
@@ -86,7 +104,7 @@ static void zathura_document_widget_get_render_range(zathura_t *zathura, unsigne
 
   int64_t max_pages = MIN(number_of_pages, cairo_max_size * priv->pages_per_row / page_height);
   *start_index = MAX(0, current_page - max_pages);
-  *page_count = MIN(number_of_pages, current_page + max_pages);
+  *page_count = MIN(number_of_pages - *start_index, current_page + max_pages);
```
jdujava, line 107: @jchilds0 Shouldn't this be
```c
*page_count = MIN(number_of_pages - *start_index, current_page - *start_index + max_pages);
```
that way `*page_count <= 2*max_pages` (even in the middle of a document).

jchilds0: Github didn't give me a notification of this review for some reason. Yes it should have been that, and the changes in #752 fixed this by making `page_count = max_pages`.

PR #742, opened 2025-05-06
Add Document open and document close d-bus signals

SebastiaanBooman:
This PR adds d-bus signals for document open and document close events.

I would like to be able listen for document open and close events. In this PR the document metadata is included in the open event, whilst the close event only contains the filename. Perhaps it would be a better idea to ensure the data sent in the close and open events are identical.

Perhaps a different/external document identifier could be used as the same file can be opened multiple times in the same Zathura process. This would make it hard to track exactly which file got opened/closed,

The first commit contains a refactor for the meta_fields struct retrieval. I added a const contraint to zathura_document_get_plugin in internal.h. This is due to having to import internal.h in document.c as it uses zathura_document_information_type_t. Without this change I got a "conflicting types" error in document.c. I am not that experienced in C so let me know if this change does not make sense.

In the future I would also like to add events for page switches, which would include:
- page number that got switched to
- previous page number
- document identifier (filename ?)

### DIFF
--- a/zathura/commands.c
+++ b/zathura/commands.c
@@ -221,39 +221,22 @@ bool cmd_info(girara_session_t* session, girara_list_t* UNUSED(argument_list)) {
     return false;
   }
 
-  struct meta_field {
-    const char* name;
-    zathura_document_information_type_t field;
-  };
+  size_t meta_fields_count = 0;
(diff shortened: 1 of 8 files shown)


sebastinas:
Sorry for the delay. I rebased and merged your changes.

PR #745, opened 2025-06-02
feat(render): optional lightness adjustment when recoloring

### LINKED ISSUE #276: Enhance font thickness
*On GitLab by @miseran on Jun 9, 2021, 10:54*

---

On low DPI displays, some fonts (like computer modern) can be very thin and hard to read. This is especially the case when using a dark recolor. It would be great if zathura could provide a feature to enhance legibility in that case.

I don't know if there is a way to improve this on the side of poppler / mupdf through font hinting or some such. If not, I suggest a setting to apply gamma correction to the output. I hacked together a proof of concept that applies a gamma correction of 1.5 before recoloring, see the following screenshot. In my opinon, the gamma corrected output (the lower screenshot) is much easier on the eyes. If there is interest, I can try implementing gamma correction as a proper setting.

zathura-gamma

MithicSpirit:
Some thinner fonts (e.g., Computer Modern Roman) can be harder to read when recoloring. This implements the option `recolor-adjust-lightness`, which adjusts the lightness before recoloring, improving readability.

The adjustment is implemented with a quartic curve. The adjusted lightness is then averaged against the original lightness, weighed by saturation, which makes colors less “washed out” (this was particularly annoying for graphics like TikZ). The weighing is not implemented when “recolor-keephue” is disabled for technical reasons, although it also improves the readability of darker colors.

Fixes #276

Note: rewritten June 6th, June 9th.

### DIFF
--- a/zathura/render.c
+++ b/zathura/render.c
@@ -39,6 +39,7 @@ typedef struct private_s {
     bool enabled;
     bool hue;
     bool reverse_video;
+    bool adjust_lightness;
   } recolor;
 
   atomic_bool about_to_close; /**< Render thread is to be freed */
@@ -93,9 +94,10 @@ static void zathura_renderer_init(ZathuraRenderer* renderer) {
   g_mutex_init(&priv->mutex);
 
   /* recolor */
-  priv->recolor.enabled       = false;
-  priv->recolor.hue           = true;
-  priv->recolor.reverse_video = false;
+  priv->recolor.enabled          = false;
+  priv->recolor.hue              = true;
+  priv->recolor.reverse_video    = false;
+  priv->recolor.adjust_lightness = false;
 
   /* page cache */
   priv->page_cache.size             = 0;
@@ -291,6 +293,21 @@ void zathura_renderer_enable_recolor_reverse_video(ZathuraRenderer* renderer, bo
   ZathuraRendererPrivate* priv = zathura_renderer_get_instance_private(renderer);
   priv->recolor.reverse_video  = enable;
 }
+
+bool zathura_renderer_recolor_adjust_lightness_enabled(ZathuraRenderer* renderer) {
+  g_return_val_if_fail(ZATHURA_IS_RENDERER(renderer), false);
+
+  ZathuraRendererPrivate* priv = zathura_renderer_get_instance_private(renderer);
+  return priv->recolor.adjust_lightness;
+}
+
+void zathura_renderer_enable_recolor_adjust_lightness(ZathuraRenderer* renderer, bool enable) {
+  g_return_if_fail(ZATHURA_IS_RENDERER(renderer));
+
+  ZathuraRendererPrivate* priv   = zathura_renderer_get_instance_private(renderer);
+  priv->recolor.adjust_lightness = enable;
+}
+
 void zathura_renderer_set_recolor_colors(ZathuraRenderer* renderer, const GdkRGBA* light, const GdkRGBA* dark) {
   g_return_if_fail(ZATHURA_IS_RENDERER(renderer));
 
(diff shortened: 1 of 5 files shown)


zathura/callbacks.c
```diff
@@ -568,6 +568,60 @@ void cb_setting_recolor_keep_reverse_video_change(girara_session_t* session, con
   }
 }
 
+void cb_setting_recolor_adjust_lightness_change(girara_session_t* session, const char* name,
```
sebastinas, line 571: All these callbacks can be moved to config.c.

zathura/render.c
```diff
@@ -518,6 +567,13 @@ static void recolor_slow(ZathuraRendererPrivate* priv, unsigned int page_width,
       rgb2.blue * rgb2.alpha - l2,
   };
 
+  bool adjust_lightness = priv->recolor.adjust_lightness;
+  bool adjust_lightness_invert = priv->recolor.adjust_lightness_invert;
+  double adjust_lightness_threshold = priv->recolor.adjust_lightness_threshold;
+  if (adjust_lightness_invert)
```
sebastinas, line 573: We are using braces also for single statement blocks.

zathura/render.c
```diff
@@ -542,6 +598,22 @@ static void recolor_slow(ZathuraRendererPrivate* priv, unsigned int page_width,
       /* compute h, s, l data   */
       double l = weights[0] * rgb[0] + weights[1] * rgb[1] + weights[2] * rgb[2];
 
+      if (adjust_lightness) {
+        /* Work with the complement when inverting. */
+        if (adjust_lightness_invert)
```
sebastinas, line 603: As above

zathura/render.c
```diff
@@ -542,6 +598,22 @@ static void recolor_slow(ZathuraRendererPrivate* priv, unsigned int page_width,
...
+          l = 1 - l;
+        /* Only apply the lightness adjustment if it is below the threshold. */
+        if (l < priv->recolor.adjust_lightness_threshold) {
+          /* Rescale the factor by the threshold so the curve is continuous. */
+          double factor = l * inv_adjust_lightness_threshold;
+          /* Apply factor twice for cubic curve. */
+          l = l * factor * factor;
+        }
+        /* Undo complement when inverting. */
+        if (adjust_lightness_invert)
```
sebastinas, line 613: As above

zathura/render.c
```diff
@@ -591,6 +663,13 @@ static void recolor_fast(ZathuraRendererPrivate* priv, unsigned int page_width,
 
   const double rgb_diff[] = {rgb2.red - rgb1.red, rgb2.green - rgb1.green, rgb2.blue - rgb1.blue};
 
+  bool adjust_lightness = priv->recolor.adjust_lightness;
+  bool adjust_lightness_invert = priv->recolor.adjust_lightness_invert;
+  double adjust_lightness_threshold = priv->recolor.adjust_lightness_threshold;
+  if (adjust_lightness_invert)
```
sebastinas, line 669: As above

sebastinas (changes requested):

MithicSpirit:
This has been updated to be significantly simpler. The threshold and inversion were removed, and now it uses a squared curve rather than cubic. However, the effect is reduced linearly as saturation increases.

I found that inversion was useless (it would only help with readability if the foreground color was darker than the background, but that is the default and already looks fine imo).

As for the threshold, that was mainly implemented to prevent the effect from making colorful graphics hard to see (the increase in lightness makes the colors appear "washed out", which blends them together). This is much better handled by scaling with saturation.

I'll be testing this for a few more days again, but I think it's pretty close to being ready for review now.

MithicSpirit:
Some more testing indicated that a quadratic curve was still insufficient. We are now up to a quartic curve, which offers a good balance of maximizing legibility without making graphics illegible (due to the saturation weighing).

I'm tempted to start adding options to control these things, but I'm trying to hold back for now. Maybe in the future a better formula can be determined.

PR #749, opened 2025-06-03
Small File Explorer

dgshanee:
zathura_fe

It's not nearly finished, but I think an :fe command to access a file explorer could be neat. 

My implementation required Typst. Let me know


dgshanee:
@sebastinas Cool. I'll start working in that direction now. Thanks!

PR #752, opened 2025-06-06
Scroll search

### LINKED ISSUE #751: Problems with dynamic page allocation #721

### Problem

I encounter various issues with the dynamic page allocation introduced in #721.

Example is the binding `map G scroll bottom`, which now (for large enough PDFs) does not scroll completely to the bottom, but only to (I guess) to the bottom of current dynamically allocated page range.

Similarly, there are problems with the searching (for large enough PDFs).

### Steps to reproduce

Just try out `map G scroll bottom` for a PDF with ~200 or more pages.

Also try to search for any phrase, then try pressing `n` multiple times, and you should encounter weird behavior. The view starts to jump to pages where there is no match (after scrolling with `j`/`k` the view sometimes gets updated to the correct page).

### Expected behavior

Scroll to bottom and searching should work as before introducing #721.

latest development build

latest development build

### zathura backend

poppler

jchilds0:
Fixes #751 

Scroll bottom doesn't update the current page number before the document widget is rendered, so we use the document position to get the current page number instead.

Search position was more difficult. When searching, if the document grid changes, the vadjustment would not have updated to reflect the current size of the grid resulting in incorrect adjustments. It seems like changing the set adjustment to update on the mainloop using `g_idle_add_once` fixes this. I'm not sure if this is the right way to solve this, or if get adjustment should also be in the mainloop, open to feedback.

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -320,8 +320,8 @@ void cb_page_layout_value_changed(girara_session_t* session, const char* name, g
   bool page_right_to_left = false;
   girara_setting_get(zathura->ui.session, "page-right-to-left", &page_right_to_left);
 
-  zathura_document_widget_set_mode(zathura, page_padding, pages_per_row, first_page_column, page_right_to_left);
   zathura_document_set_page_layout(zathura_get_document(zathura), page_padding, pages_per_row, first_page_column);
+  zathura_document_widget_set_mode(zathura, page_padding, pages_per_row, first_page_column, page_right_to_left);
 }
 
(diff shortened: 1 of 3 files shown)


jdujava:
Just applying 927d314f9e0a09da8a8a19b3c4a5c22b88700483 almost fixes the `scroll bottom` issue, but (for long enough document) after doing `gg` and `G` (mapped to `scroll bottom`) I can still scroll a tiny bit down.

When I also include b42905b9f1b492fcd7353bbecfb18e0732e4dc65, I encounter very weird behavior, `gg` doesn't go entirely to the top when starting from the bottom, and zooming in/out is accompanied with flickering.

Issue #758, opened 2025-06-29
Horizontal/Vertical page padding


0xdeadbeer:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem

Would it be feasible to add two distinct options to control horizontal/vertical padding between pages so I can have wider gaps in dual mode in horizontal than in vertical? 

### Expected behavior

Something like this: 
`set page-h-padding 10`
`set page-v-padding 10` 

As well as support this for compatibility:
`set page-padding 10`

PR #759, opened 2025-06-30
Break page-padding into two separate options

0xdeadbeer:
Replace page-padding with two separate options for horizontal and vertical padding. 

Issue: #758 

Example: 
`set page-v-padding 10` 
`set page-h-padding 200`


.gitignore
```diff
@@ -8,6 +8,7 @@
 *.sw[a-z]
 *.pyc
 tags
+.cache/
```
sebastinas, line 11: This is unrelated. Please drop this change.

sebastinas (changes requested):

PR #763, opened 2025-07-13
zoom shortcuts for document index

### LINKED ISSUE #754: index mode zooming

### Problem

Key bindings of zooming don't work in index mode. This is a problem when the font is too small for a long and dense index.

### Expected behavior

Zooming keybinds (`+-=` etc) should work in index mode.

jchilds0:
Resolves #754.


jchilds0:
Yes, I didn't know about the `font` option. That's a better way to do it

PR #768, opened 2025-07-18
fix page grid attaching

### LINKED ISSUE #766: Even/odd pages in dual-page layout not displayed correctly
When viewing document in dual-page layout (toggled by default with `d`), the odd/even pages should always be on left/right (or vice versa). What however happens is that maybe at the start (page < 60) you always have pages with even page number on the right, after you go to the bottom (for example with `G`), you can find the opposite (if the document has >150 or something pages).

After slowly scrolling up to find the page where the jump happened, one can see something like this:
Image
After scrolling even further, the pages rearrange. I believe this issue was introduced (or is connected) with #721.

jchilds0:
resolves #766 

Grid attaching of pages didn't correctly account for the first page column. 
Modified the document widget to render a number of page rows, and every page in the given rows.


jchilds0:
Actually I think this needs some more work, I'll close for now.

PR #769, opened 2025-07-18
Add vim's expand/collapse fold bindings to index mode

druckdev:
The tree in the index mode is comparable to folds in vim, thus its mappings around opening and closing folds make sense here too. Add the following aliases:

- `zo` to expand
- `zc` to collapse
- `zR` to expand all
- `zM` to collapse all

and the following bindings to new functions:

- `zO` to expand recursively
- `zC` to collapse recursively

Additionally, do some minor reshuffling of the collapse* and expand* entries in the touched code for a more consistent order, document the other previously missing default bindings in index mode and remove the duplicate shortcut `j`.


zathura/shortcuts.c
```diff
@@ -1110,10 +1104,27 @@ bool sc_navigate_index(girara_session_t* session, girara_argument_t* argument, g
       gtk_tree_path_down(path);
     }
     break;
+  case EXPAND_RECURSIVE:
+    if (gtk_tree_view_expand_row(tree_view, path, TRUE)) {
+      gtk_tree_path_down(path);
+    }
+    break;
```
druckdev, line 1111: The cases `EXPAND` and `EXPAND_RECURSIVE` could also be put together:

```suggestion
  case EXPAND:
  case EXPAND_RECURSIVE:
    if (gtk_tree_view_expand_row(tree_view, path, argument->n == EXPAND_RECURSIVE)) {
      gtk_tree_path_down(path);
    }
    break;
```

What do you think?

PR #772, opened 2025-07-29
don't focus the tree view in index mode

### LINKED ISSUE #709: [Bug]: Arrow keys not working in nested table of contents

- [X] I have searched the existing issues

### Problem

Hi All, 
I encountered what seems to be a bug today.
Simply put, navigating in nested table of contents (of any pdf file of choice) used to work just using Arrow Keys (until version 0.5.8-1). 
Now I can open a nested toc but I'm unable, using keyboard only, to navigate to nested sections inside it.

### Steps to reproduce

1- Launch zathura
2- Open a pdf file with a nested Table of Contents, for example [https://www.math.ias.edu/~lurie/papers/HA.pdf](url) .
3- Open the table of contents using Tab key,
4- Try to navigate, using keyboard only, to nested sections in the toc, like 3.1.2

### Expected behavior

Expected behavior: the navigation should work using Up/Down/Left/Right arrow keys (right and left to open and close, respectively, nested sections). 

0.5.11

0.4.5

### zathura backend

pdf-mupdf

jchilds0:
Resolves #709 .

Its not entirely clear why the tree_view grabbed focus in #677 , but this meant the tree view would consume inputs like arrow keys and by default up and down move in a tree view and left right do nothing.
This meant `sc_navigate_index` would not be called on arrow inputs.

In any case, I think its ok to disable the grab while search is still in development, and I think the key press should come from girara as with other inputs, rather than calling grab focus on the tree view.

### DIFF
--- a/zathura/shortcuts.c
+++ b/zathura/shortcuts.c
@@ -1252,8 +1252,8 @@ bool sc_toggle_index(girara_session_t* session, girara_argument_t* UNUSED(argume
     }
 
     girara_set_view(session, zathura->ui.index);
-    GtkTreeView* tree_view = gtk_container_get_children(GTK_CONTAINER(zathura->ui.index))->data;
-    gtk_widget_grab_focus(GTK_WIDGET(tree_view));
+    // GtkTreeView* tree_view = gtk_container_get_children(GTK_CONTAINER(zathura->ui.index))->data;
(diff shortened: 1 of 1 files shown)


TethysSvensson:
Would you mind doing a new release with this?

PR #780, opened 2025-08-12
Add utility plugin support and TTS API functions

ubuntupunk:
This PR adds utility plugin support to Zathura, enabling plugins to register utility functions and providing TTS API functions for text-to-speech functionality.


sebastinas:
It is not really clear to me where you want to go with this. Could you please give some context what you want achieve.

ubuntupunk:
Thanks for responding. I started out attempting to create a Zathura TTS plugin. Then realised it would be impossible to solve the initiatilisation without modifiying Zathura to allow utility plugins since the current functions are essentially looking for document types, and the result would be very inelegant and hacky. I have a rather sketchy plugin that works using piper, and have found a better method to avoid this without requireing a dependency, when I have time I will come back to the project [zathura tts plugin](https://github.com/ubuntupunk/zathura-liberated)

sebastinas:
Wouldn't it be easier to leverage the support that Gtk already has for accessibility? In the end, we'd just need to implement the `GtkAccessible` interfaces for the page widget (with potentially a function to get the text of the full page from the plugin).

ubuntupunk:
Zathura is primarily a CLI app for terminal-centric users or those trapped inside headless systems. If you want to create another layer of complexity on the desktop, or support a GTK interface for a GTK wrapper, then sure, but the utility function is a clean solution that could potentially enable *other utilities as opposed to document types. For example, signing PDF documents by interacting with GPG via a gpg plugin etc. I get why this PR raises such questions (why don't we support QT and every other framework out there?)

ubuntupunk:
That's funny, 'Zathura’s UI is built on girara, which provides the lightweight, vi-like interface (document view, status bar, input bar). Girara is tightly coupled with GTK for rendering, meaning it cannot operate in a text-only environment.' Been using it in my terminal for years. My apologies for the confusion.

Issue #686, opened 2024-09-30
Presentation mode shows previous and next slide


Irubataru:
If I make a presentation in beamer and the aspect ratio of the presentation isn't the same as the monitor it is displayed on, then presentation mode in zathura shows the slide (page) before and after at the top and bottom. I'd rather that it showed black bars. This would also make sure it was centered when showing the first and last slide.

Probably related to #103 

How it looks now:
![image](https://github.com/user-attachments/assets/96422774-ac19-4089-9a1d-fa9038d216ea)

How I expect it to look:
![image](https://github.com/user-attachments/assets/bee6f571-e029-4b27-99b9-97a314f351af)

Zathura version:
```
zathura 0.5.8
girara 0.4.4 (runtime: 0.4.4)
(plugin) ps (0.2.8) (/usr/lib/zathura/libps.so)
(plugin) pdf-mupdf (0.4.4) (/usr/lib/zathura/libpdf-mupdf.so)
(plugin) djvu (0.2.10) (/usr/lib/zathura/libdjvu.so)
```

tobsto:
I would like to second that. Such a feature would be very much appreciated. It is for example the standard way in evince.

kuon:
Yes being able to use zathura for presentations would be very appreciable.

Basically only one page should be displayed at a time, and page-up/down should move to previous/next page without moving the page frame.

ehula:
This feature is desperately needed.

PR #833, opened 2025-11-22
page padding updates

jchilds0:
Split document get padding into vertical and horizontal components.
Handle page padding in page_number_to_position. This resolves an issue brought up in #686, where navigation drifts in presentation mode if page padding is increased to have a single page visible.


jchilds0:
I've tested navigate and search on a couple of documents with page padding default of 1 and 300, and it seems to work as expected but if others can check for any issues introduced that'd be good.

Issue #771, opened 2025-07-28
Rotate pages independently


3nt3:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem

You can only rotate all pages in the document.

### Expected behavior

Proposed shortcut `<alt-r>/<alt-R>`. For documents which have been created by people who did not rotate the pages correctly themselves or for scanned documents with inconsistent page rotations.

PR #838, opened 2025-12-03
feat: implement per-page rotation logic

Aadeshveer:
Relates to Issue #771 

### Summary
This PR implements individual page rotation support. Currently, Zathura only supports rotating the entire document. This adds the ability to rotate specific pages (e.g., for scanned documents with mixed orientation) using `Alt+r` (CW)

### Implementation Details
*   **Data Structure:** Added a `rotate` attribute to `zathura_page_s` (initialized to 0) along with getter/setter functions in `page.c`.
*   **Rendering:** Modified `zathura_page_widget_draw` (in `page-widget.c`) to sum the document rotation and individual page rotation before passing the angle to Cairo.
*   **Logic:** Added `individual_page_calc_height_width` in `adjustment.c` to handle scale calculations for rotated pages.
*   **Shortcuts:** Added `page_rotate` function in `shortcuts.c` and mapped it to `<Alt>r` in `config.c`.

### The Layout Trade-off (Request For Comments)
The current Zathura grid layout assumes uniform cell sizes for efficient scrolling. Rotating a single page (e.g., Portrait -> Landscape) creates a mismatch between the page widget size and the allocated grid cell, leading to scrolling glitches.

To resolve this without rewriting the entire grid virtualization, I implemented a "Bounding Box" approach in `zathura.c`:
```c
*width  = MAX(*width, *height);
*height = MAX(*height, *width);
```
This forces the layout cells to be square (based on the largest dimension).
*   **Pro:** Rotation works perfectly without clipping or scroll breaking.
*   **Con:** In "Fit Width" mode, portrait pages will have horizontal padding to accommodate the potential rotation.

I believe this is a valid trade-off for the feature, but I am open to suggestions on handling the variable row height if strictly necessary.

### Testing
1.  Open a PDF.
2.  Press `Alt+r` to rotate the current page.
3.  Press `r` to verify document-level rotation still works on top of page rotation.

### Future Work / Limitations
*   **Persistence:** Currently, individual page rotations are **not** saved to the database (sqlite) and will reset upon closing the application. I believe adding database schema changes/serialization for per-page state should be handled in a separate, follow-up PR to keep this initial implementation focused on the rendering and logic layer.


jchilds0:
Using square bounding box only accounts for rotations in 90 deg increments, if you rotate by 45 deg for example the width will be more than either height or width of the original page individually. 

That's ok though, I'm working on an updated document widget which will have variable page sizing. It'll call `page_calc_width_height` on each page and give it atleast the size it reports there. The page_calc function you've written needs updating to handle arbitrary rotation angles.

Aadeshveer:
@jchilds0 Thanks!
Since we are constrained to 90-degree increments for now, the square bounding box approach (max(w,h)) ensures the layout remains stable and scrollable without clipping, even if it adds some horizontal whitespace in "Fit Width" mode.
I think this is a solid improvement for the current codebase as it enables a highly requested feature without waiting for the larger widget refactor. I'd love to see this merged as an intermediate solution, and I'd be happy to help update/refactor it once your new document widget lands!
I've marked the PR as ready for review.

PR #851, opened 2026-01-05
document widget: features

### LINKED ISSUE #103: [Feature Request] Display single pages
*On GitLab by @Skillmon on Jan 15, 2019, 17:25*

---

Would be nice to have an option to disable continuous scrolling in normal mode. (Sorry if that already exists, I couldn't find it)

### LINKED ISSUE #289: Zathura's best-fit command ('a') for documents with varying page size
*On GitLab by @strongthong on Sep 2, 2021, 02:18*

---

Someone has written about this almost 3 years ago, but I don't think it's actually been brought up given I still see the problem persisting myself.

I'll copy it verbatim: https://unix.stackexchange.com/questions/468875/zathuras-best-fit-command-a-for-documents-with-varying-page-size

> I'm usually using the wonderful document viewer 'zathura' for reading PDF-files. Typically, for some documents, the first page of the document is slightly larger than the rest (for example, a photograph of a book cover). In this case the 'best-fit' function (key 'a') of zathura always adjusts the zoom relative to the first page of the document, implying that for the rest of the document the pages are displayed smaller than they should be. This is particularly annoying when reading documents in 'double-page mode', leading to results as terrible as the following:

yKsUd 
>**My Question:** Is there any way to change this behaviour such that the 'best-fit' function adjusts the size relative to the currently viewed page?

### LINKED ISSUE #447: Fit width mode for variable page sizes

### LINKED ISSUE #706: Border between pages in the two pages view

- [X] I have searched the existing issues

### Problem

When I open a book in pdf i see a space between two pages. How to remove that space and get that back as before? 

screenshot_1735743514

### Steps to reproduce

Open pdf with many pages. Press 'd' to display two pages side by side. Results: space between pages.

Same on mupdf backend.

### Expected behavior

No space between pages.

zathura 0.5.11

girara 0.4.5 (runtime: 0.4.5)

### zathura backend

(plugin) pdf-poppler (0.3.3) (/usr/lib/zathura/libpdf-poppler.so)

### LINKED ISSUE #724: Fit Width Zoom Mode for Pages Individually

### Problem

I have been using Zathura for pretty long time now.
The only gripe with it is the fit width zoom behavior.

It only does the zooming based on the page with the maximum width not with the width of every page individually.
Okular does what I think is the best implementation of **Fit Width**. It does on per-page basis making it consistent across the document, not making the user to change the zoom in for the pages with normal width and zoom out for the page with maximum width.

**_Here is how Zathura does fit width zooming:_**

Image

**_And here is how Okular does fit width zooming:_**

Image

### Expected behavior

If it is possible to make zathura do this please make it happen as it will surely be better than the current behavior.

Thanks a lot for making this awesome program.

jchilds0:
Resolves #289, resolves #447, resolves #724, resolves #706, resolves #103. 

https://github.com/pwmt/zathura/pull/851/commits/6a90d1cb4d83504921700ff326e0e5985cbace1a : adds a single page mode, which resolves #103 .
https://github.com/pwmt/zathura/pull/851/commits/9f2088a99d22f99e30139191414459f4a1837fe3 : adds per page zoom and a shortcut to change the zoom for the current page, currently the zoom value is only stored in the page object and is thus lost when changing documents. Each page is allocated `page_width * page_zoom * document_zoom`, similarly for height.
https://github.com/pwmt/zathura/pull/851/commits/fe937216869aa700ef2692a4f530b96f0c0be08a : implements two modes, "equal_width" and "equal_height", which sets the zoom for each page so the width and height respectively are the same across all pages. This can be combined with adjust modes such as best-fit to have fit width for variable pages. I believe this covers all the use cases described in #289, #449 and #724.

I'll leave this in draft mode until #849 is merged.

### DIFF
--- a/zathura/adjustment.h
+++ b/zathura/adjustment.h
@@ -12,7 +12,7 @@
  * desired.
  *
  * @param document the document
- * @param page the page 
+ * @param page the page
  * @param page_height the scaled and rotated height
  * @param page_width the scaled and rotated width
  * @param rotate honor page's rotation
--- a/zathura/document-widget.c
+++ b/zathura/document-widget.c
@@ -74,16 +74,15 @@ static void zathura_document_widget_class_init(ZathuraDocumentWidgetClass* class
       g_param_spec_pointer("zathura", "zathura", "the zathura instance",
                            G_PARAM_WRITABLE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS));
 
-  g_object_class_install_property(
-      object_class, PROP_LAYOUT_MODE,
-      g_param_spec_int("layout-mode", "layout-mode", "set the page layout mode", 
-                       0, DOCUMENT_WIDGET_MODE_COUNT, DOCUMENT_WIDGET_GRID, 
-                       G_PARAM_WRITABLE | G_PARAM_READABLE));
+  g_object_class_install_property(object_class, PROP_LAYOUT_MODE,
+                                  g_param_spec_int("layout-mode", "layout-mode", "set the page layout mode", 0,
+                                                   DOCUMENT_WIDGET_MODE_COUNT, DOCUMENT_WIDGET_GRID,
+                                                   G_PARAM_WRITABLE | G_PARAM_READABLE));
 
-  g_object_class_install_property(
-      object_class, PROP_PAGES_RIGHT_TO_LEFT,
-      g_param_spec_boolean("pages-right-to-left", "pages-right-to-left", "layout pages left to right", false, 
-                           G_PARAM_WRITABLE | G_PARAM_READABLE));
+  g_object_class_install_property(object_class, PROP_PAGES_RIGHT_TO_LEFT,
+                                  g_param_spec_boolean("pages-right-to-left", "pages-right-to-left",
+                                                       "layout pages left to right", false,
+                                                       G_PARAM_WRITABLE | G_PARAM_READABLE));
 
   g_object_class_override_property(object_class, PROP_HADJUSTMENT, "hadjustment");
   g_object_class_override_property(object_class, PROP_VADJUSTMENT, "vadjustment");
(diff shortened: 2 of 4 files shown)


sebastinas:
I have pushed the rebased version to `jchilds0-single-page`.

While testing I noticed that movement with hjkl does not work when in single page mode and zoomed in.

jchilds0:
> I have pushed the rebased version to `jchilds0-single-page`.
> 
> While testing I noticed that movement with hjkl does not work when in single page mode and zoomed in.

I didn't consider that case, the current page is just given the whole screen to draw to. I'll add that in

jchilds0:
Fixed hjkl in single page mode and added shortcut documentation. Currently in single page mode, if the document is smaller than the width/height it gets centered, but when its zoomed in more then its positioned as though the whole grid is visible. This makes it a little weird when zooming as there can be jumps in position going from centered to document position. Do we want a different behavior?

sebastinas:
Looks good.

It looks as if vertical scrolling is limited to the page boundaries (when you reach the right or left bound of the page, scrolling with h/l does not go any further). But for horizontal scrolling this is different. One can scroll beyond the top and bottom of the page.

PR #901, opened 2026-03-27
Added whitespace in front of "Search: ..."

unrelated-stardust:
I think this formatting
image
is better than this
image


sebastinas:
No, we'll want some padding between the statusbar items.

sebastinas:
The widget class was set incorrectly. This should now have spacing.

PR #910, opened 2026-04-15
fix various issues

valoq:
Document info text (title, author) can contain special characters that mess up the notification display. The same issue exists for link targets and link copy.

Fixed by added escaping

Also added a check for large page sizes that would break rendering

### DIFF
--- a/zathura/page.c
+++ b/zathura/page.c
@@ -137,7 +137,7 @@ void zathura_page_set_width(zathura_page_t* page, double width) {
     return;
   }
 
-  if (!isfinite(width) || width <= 0.0 || width > 1e7) {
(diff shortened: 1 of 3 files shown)


zathura/page.c
```diff
@@ -136,6 +137,11 @@ void zathura_page_set_width(zathura_page_t* page, double width) {
     return;
   }
 
+  if (!isfinite(width) || width <= 0.0 || width > 1e7) {
```
sebastinas: Where is the upper bound coming from?

valoq: Sorry that was left over from testing the boundary

valoq:
I added bookmark escaping too. This should cover all relevant cases

PR #911, opened 2026-04-15
fix loading message

### LINKED ISSUE #864: "Loading ..." display while scrolling

### Problem

When scrolling a document with either the arror keys or the mouse wheel, the current version of zathura displays a "loading ..." message in the middle of the document for a split second.
Going throught eh entire document once seems to fix the issue for soem reason.

### Steps to reproduce

Open a pdf document, use arro keys or mouse wheel to scroll through it

### Expected behavior

No loading message on the first pages. The Messaage should probably only be displayed when navigating to a distant page or going quickly through many pages.

2026.01.30

2026.01.30

### zathura backend

poppler

valoq:
fixes #864

~~To avoid showing the loading message, this patch just waits 500ms before showing the "Loading..." text. During that wait it just shows a blank page. Since most pages finish rendering well before 500ms, you never see the text during normal scrolling. If a page genuinely takes a long time to render, the text still shows up after half a second~~

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -63,8 +63,17 @@ void update_visible_pages(zathura_t* zathura) {
       /* make page visible */
       if (zathura_page_get_visibility(page) == false) {
         zathura_page_set_visibility(page, true);
-        zathura_page_widget_update_view_time(zathura_page_widget);
         zathura_renderer_page_cache_add(zathura->sync.render_thread, zathura_page_get_index(page));
+      }
+      zathura_page_widget_update_view_time(zathura_page_widget);
+      /* keep adjacent pages rendered so scrolling lands on ready content */
+      if (page_id > 0) {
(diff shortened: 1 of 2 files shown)


valoq:
After digging a lot more around this issue, Im pretty sure I found the actual cause:

Commit 9035e15b replaced the GtkGrid page layout with a tighter custom container. The old grid caused GTK to draw pages beyond the visible area and triggered their renders via the draw callback. The new layout only draws visible pages, so when the user scrolls to a page for the first time, its render hasn't started and the "Loading..." screen shows until the render completes

zathura_page_widget_update_view_time now queues a render if the widget has no surface. update_visible_pages calls it on each visible page and on the pages immediately before and after. Renders for adjacent pages are in flight before the user scrolls there

jchilds0:
If you have `pages-per-row` set to more than 1, this will only pre-render the first page of the row.

PR #913, opened 2026-04-15
feat: confirm open link

valoq:
zathura is one of the few cases where links still open to external resources without confirmation dialog. Since this is a common source for phishing and social engineering, most applications now include a confirmation. This adds a simple dialog that shows the url to be opened (or the external program) and waits for user confirmation with enter or 'y'

### DIFF
--- a/zathura/links.c
+++ b/zathura/links.c
@@ -239,12 +239,11 @@ static gboolean cb_link_confirm(GtkEntry* entry, void* data) {
     return true;
   }
 
-  char* input = gtk_editable_get_chars(GTK_EDITABLE(entry), 0, -1);
+  g_autofree char* input = gtk_editable_get_chars(GTK_EDITABLE(entry), 0, -1);
 
-  /* Accept: empty string (bare Enter), "y", or "Y" */
+  /* Accept: empty string (bare Enter), or confirmation string */
   const bool confirmed =
-      (input == NULL || input[0] == '\0' || (input[1] == '\0' && (input[0] == 'y' || input[0] == 'Y')));
-  g_free(input);
+      (input == NULL || input[0] == '\0' || g_strcmp0(input, _("y")) == 0 || g_strcmp0(input, _("Y")) == 0);
 
   if (confirmed) {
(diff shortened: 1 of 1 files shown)


zathura/links.c
```diff
@@ -211,6 +211,95 @@ static void link_launch(zathura_t* zathura, const zathura_link_t* link) {
...
+  if (entry == NULL || ctx == NULL) {
+    link_confirm_data_free(ctx);
+    return true;
+  }
+
+  char* input = gtk_editable_get_chars(GTK_EDITABLE(entry), 0, -1);
+
+  /* Accept: empty string (bare Enter), "y", or "Y" */
+  const bool confirmed =
+      (input == NULL || input[0] == '\0' || (input[1] == '\0' && (input[0] == 'y' || input[0] == 'Y')));
```
sebastinas: `y` / `>` could be translated. So this would need to check for equality with `_("y")` and `_("Y")`.

zathura/links.c
```diff
@@ -211,6 +211,95 @@ static void link_launch(zathura_t* zathura, const zathura_link_t* link) {
...
+ * Opens the link only if the user pressed Enter with empty input or 'y'
+ */
+static gboolean cb_link_confirm(GtkEntry* entry, void* data) {
+  link_confirm_data_t* ctx = data;
+  if (entry == NULL || ctx == NULL) {
+    link_confirm_data_free(ctx);
+    return true;
+  }
+
+  char* input = gtk_editable_get_chars(GTK_EDITABLE(entry), 0, -1);
```
sebastinas: Can be marked as `g_autofree`.

sebastinas (changes requested):

valoq:
Done

antinomie8:
Is it possible to add an option for toggling this behavior ? When I click e.g. a web url link in a pdf I don't necessarily expect a confirmation dialog to open before running `xdg-open`. Currently it looks like there's no choice but stick to an older version of zathura if one wants to keep the previous behavior for opening links.

PR #914, opened 2026-04-16
feat: Can use + and - in pagejumps using :page

Sai-Prabhav:
This is a small feature that I personally found missing. I am not a great at c so please let me know if there is any mistake/improvements to be made in the code. Also this is my first time contributing let me know I there is some part of process that I missed.

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -579,13 +579,32 @@ bool cb_unknown_command(girara_session_t* session, const char* input) {
   /* check for number */
   const size_t size = strlen(input);
   for (size_t i = 0; i < size; i++) {
-    if (g_ascii_isdigit(input[i]) == FALSE) {
+    if (g_ascii_isdigit(input[i]) == FALSE && input[i] != '-' && input[i] != '+') {
       return false;
     }
   }
 
+  int jump_page = 0;
+  int a         = 0;
+  int sign      = 1;
+  for (size_t i = 0; i < size; i++) {
+    if (input[i] == '-') {
+      jump_page += sign * a;
+      a    = 0;
+      sign = -1;
+    } else if (input[i] == '+') {
+      jump_page += sign * a;
+      a    = 0;
+      sign = 1;
+    } else {
+      a = a * 10 + input[i] - '0';
+    }
+  }
+
+  jump_page += sign * a;
+
   zathura_jumplist_add(zathura);
-  page_set(zathura, atoi(input) - 1);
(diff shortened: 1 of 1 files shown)


zathura/callbacks.c
```diff
@@ -579,13 +579,32 @@ bool cb_unknown_command(girara_session_t* session, const char* input) {
   /* check for number */
   const size_t size = strlen(input);
   for (size_t i = 0; i < size; i++) {
-    if (g_ascii_isdigit(input[i]) == FALSE) {
+    if (g_ascii_isdigit(input[i]) == FALSE && input[i] != '-' && input[i] != '+') {
```
sebastinas, line 582: The signs + and - can only appear as input[0], not anywhere else.

zathura/callbacks.c
```diff
@@ -579,13 +579,32 @@ bool cb_unknown_command(girara_session_t* session, const char* input) {
...
+    } else if (input[i] == '+') {
+      jump_page += sign * a;
+      a    = 0;
+      sign = 1;
+    } else {
+      a = a * 10 + input[i] - '0';
+    }
+  }
+
+  jump_page += sign * a;
```
sebastinas, line 604: This can be simplified to read the sign first if input[0] is + or - and then keep the `atoi` call for the remaining part.

sebastinas (changes requested):

zathura/callbacks.c
Sai-Prabhav: I wanted the ability to use `:10+200-30` which results in moving to page `180`.

sebastinas: What for?

Sai-Prabhav: This might be a niche use case but this is what I use it for. Some PDF dont have the page numbers aligned so the 100th page in the book might be in 110th position in pdf so when I want to got to say 35th page in the book I can do `:110-100+35` to just jump to the page 35 in the book which is 45th position in the pdf. Also I just felt doing `:30+10` should work but its just my opinion.

Sorry if this sounds confusing don't know how to put it better.

sebastinas: If the page numbers do not align, we have the `offset` feature.

Sai-Prabhav: I was not aware of that thanks for letting me know. I personally like the feature. Also I can make what you just said in comments if starting with + or - then it will be relative offsets or else absolutely page number

PR #919, opened 2026-05-02
feat: keep the highlighting even when falling of the view

### LINKED ISSUE #693: keep text highlighted
Hello, I would like that zathura had an option to keep selected text highlighted, even if the user has moved the mouse elsewhere. Mainly because it would be useful for screenshoting. You could have a key (escape for instance) that explicitly removes the highlight.

nsavvide:
[UPDATED] added a `selection-keep-highlight` configuration option

When the cursor leaves the pdf page, the selection gets cleared. However for some (see issue #693 ) this behavior is unwanted. Simply adding the configuration seting  and clearing all highlights on a left click seems to be the logical change to me.

image

@sebastinas  feel free to review this code please :) This would be my first contribution so please feel free to educate me where necessary XD

closes #693


sebastinas:
Rebased and merged.

PR #924, opened 2026-05-06
fix various bugs and minor issues

breadbeard95:
fix various bugs, leaks and typos

- synctex path parsing
- /proc/self/fd handling
- pages-per-row input validation
- zoom division-by-zero
- index navigation: leak and uninitialised frees
- document grid layout under allocation failure
- print surface size limits
- swapped uri/password in synctex document open
- duplicate bookmark rows on database errors
- marks and print page numbering
- window-icon at startup
- search, dialog, recolor, export and shutdown leaks
- error checks in database, dbus and document hashing
- comment and log-message typos

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -471,16 +471,13 @@ void cb_file_monitor(ZathuraFileMonitor* monitor, girara_session_t* session) {
   g_main_context_invoke(NULL, file_monitor_reload, session);
 }
 
-static void cb_password_dialog_hide(GtkWidget* UNUSED(w), void* data);
-
 static void password_dialog_info_free(zathura_password_dialog_info_t* dialog) {
   if (dialog == NULL) {
     return;
   }
-  if (dialog->zathura != NULL && dialog->zathura->ui.session != NULL &&
-      dialog->zathura->ui.session->gtk.inputbar_dialog != NULL) {
-    g_signal_handlers_disconnect_by_func(dialog->zathura->ui.session->gtk.inputbar_dialog, cb_password_dialog_hide,
-                                         dialog);
+  if (dialog->hide_handler != 0) {
+    g_signal_handler_disconnect(dialog->zathura->ui.session->gtk.inputbar_dialog, dialog->hide_handler);
+    dialog->hide_handler = 0;
   }
   g_free(dialog->path);
   g_free(dialog->uri);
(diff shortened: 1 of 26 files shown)


zathura/database-sqlite.c
```diff
@@ -335,7 +337,9 @@ static void sqlite_db_check_layout(sqlite3* session, const int database_version,
     }
   }
   if (database_version < 2) {
-    if (sqlite3_exec(session, SQL_FILEINFO_ALTER6, NULL, 0, NULL) != SQLITE_OK) {
+    bool has_sha = false;
+    if (check_column(session, "fileinfo", "sha256", &has_sha) == false ||
+        (has_sha == false && sqlite3_exec(session, SQL_FILEINFO_ALTER6, NULL, 0, NULL) != SQLITE_OK)) {
```
sebastinas: As soon of version 1 of the database layout, we know which fields are missing. So if `database_version == 1`, we know that `fileinfo` and `sha256` need to be created. The calls to `check_column` in the block before are to check the state of the database before we introduced versions.

zathura/callbacks.c
```diff
@@ -44,7 +44,7 @@ void cb_buffer_changed(girara_session_t* session) {
   char* buffer = girara_buffer_get(session);
   if (buffer != NULL) {
     girara_statusbar_item_set_text(session, zathura->ui.statusbar.buffer, buffer);
-    free(buffer);
+    g_free(buffer);
```
sebastinas: This could be turned into a `g_autofree char* buffer ...` above.

zathura/page.c
```diff
@@ -71,9 +71,8 @@ zathura_page_t* zathura_page_new(zathura_document_t* document, unsigned int inde
     }
 
     if (page->label != NULL) {
-      char page_number_string[G_ASCII_DTOSTR_BUF_SIZE];
-      g_ascii_dtostr(page_number_string, G_ASCII_DTOSTR_BUF_SIZE, index + 1);
-      page->label_is_number = strcmp(page->label, page_number_string) == 0;
+      g_autofree char* page_number_string = g_strdup_printf("%u", index + 1);
```
sebastinas: Why this change?

zathura/callbacks.c
```diff
@@ -302,13 +302,13 @@ void cb_page_layout_value_changed(girara_session_t* session, const char* name, g
   g_return_if_fail(session->global.data != NULL);
   zathura_t* zathura = session->global.data;
 
-  /* pages-per-row must not be 0 */
+  /* pages-per-row is INT; reject non-positive to prevent uint reinterpret. */
   if (g_strcmp0(name, "pages-per-row") == 0) {
     unsigned int pages_per_row = *((const unsigned int*)value);
-    if (pages_per_row == 0) {
+    if ((int)pages_per_row <= 0) {
```
sebastinas: This covers only one case where we assume that values from settings are unsigned while they are signed internally. Please drop this change from this PR.

zathura/callbacks.c
```diff
@@ -471,10 +471,38 @@ void cb_file_monitor(ZathuraFileMonitor* monitor, girara_session_t* session) {
...
+  g_free(dialog->path);
+  g_free(dialog->uri);
+  g_free(dialog);
+}
+
+static void cb_password_dialog_hide(GtkWidget* UNUSED(w), void* data) {
+  password_dialog_info_free(data);
+}
+
+void password_dialog_arm_hide(void* data) {
```
sebastinas: This function is only used in `callbacks.c` and can be `static`.

PR #927, opened 2026-05-12
feat: add signed releases

valoq:
This adds github releases and adds sigstore signing to new releases.
https://www.sigstore.dev/

tldr:
Sigstore enables automated signing through githubs ci without the need for persistent key management. Keys are short lived and bound to the authentication process. Verification is handled through transparency logs and can be done in a single step:

```
  cosign verify-blob \
    --certificate-identity \
      https://github.com/pwmt/zathura/.github/workflows/release.yaml@refs/tags/<TAG> \
    --certificate-oidc-issuer https://token.actions.githubusercontent.com \
    --bundle zathura-<VERSION>.tar.xz.sigstore.json \
    zathura-<VERSION>.tar.xz
```

Most importantly, it causes no additional work during the release process

@sebastinas I am curious about your take on this approach and sigstore in general.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,68 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
+
+permissions:
+  contents: read
+
+jobs:
+  release:
+    name: Build, sign, and publish release artifacts
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/pwmt/github-actions-debian:forky
+      credentials:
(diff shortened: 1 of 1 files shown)


sebastinas:
We do not use github to produce releases and also do not intend to change it. The primary source for releases is pwmt.org. If there is interest in signed releases, we will sign this with a gpg key that is well connected in the web of trust.

valoq:
pgp signed releases would be a very nice addition, sigstore is just easier for most to setup, hence the suggestion.
Closing this

sebastinas:
Signatures are now available on the website.

Issue #928, opened 2026-05-12
zathura-sandbox: killed by seccomp


ilf:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem

zathura-sandbox 2026.05.11 is killed by seccomp:

```
kernel: audit: type=1326 audit(1778596192.429:8): auid=1000 uid=1000 gid=1000 ses=1 pid=61834 comm="gly-global-exec" exe="/usr/bin/zathura-sandbox" sig=31 arch=c000003e syscall=62 compat=0 ip=0x7f299e03eafb code=0x80000000
```

### Steps to reproduce

Linux: 7.0.5
glibc: 2.43+r22

### Expected behavior

no crash

### zathura version (zathura --version)

2026.05.11

### girara version (zathura --version)

2026.02.04

### zathura backend

poppler

valoq:
@ilf Can you provide any details about the trigger or your environment?
Desktop environment, Window manager, exact command or option that triggered the crash. I assume Arch is the distro? Any details may help. Was this triggered by a specific file and if so, can you provide it?

Note:
syscall=62 is the `kill` syscall

tax79:
I have the same behavior.

Arch Linux
xLibre
dwm
zathura-pdf-mupdf backend.

Command:
zathura-sandbox example.pdf

Precisely in 30 seconds the window closes with the following error:
Bad system call   example.pdf

Effected versions:
zathura 2026.05.10
zathura 2026.05.11

Non-effected versions:
zathura 2026.03.27 and earlier.

valoq:
I think I found the issue being related to X11.
If anyone has this problem without using X11 window manager, please let me know.

Note that running the sandbox on X11 provides weak opportunistic security at best since sandbox escape is very trivial

valoq:
@ilf @tax79 Could you test the fix in #929 please? it should fix the sandbox for X11

Unfortunately the X11 unit test didn't catch this one

tax79:
@valoq Didn't help.

valoq:
I tested the patch in #929 with i3, dwm, openbox and xfce without finding any additional issues.

@tax79 Can you check your audit logs please?
When you have the audit demon enabled and run into the issue, you can find the log entry with `ausearch -x zathura`

tax79:
@valoq 
journalctl -b
[...]

PR #929, opened 2026-05-12
fix zathura-sandbox

valoq:
This fixes the sandbox regression reported in #928

@sebastinas Mayby worth a new release, since zathura-sandbox will always crash after 30 seconds in 2025.05.11


ilf:
Thank you. Can we get a release with this?

valoq:
> Thank you. Can we get a release with this?

Up to @sebastinas 
Since the next release may take a while for the gtk4 port, it would be nice to have a new release in the meantime.

sebastinas:
I'll do a release today or tomorrow.

Issue #907, opened 2026-04-11
--fork causes error spam on MacOS


madskjeldgaard:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem

Hi! 

When using zathura on MacOS and invoking the --fork flag, my terminal gets spammed by these errors:

``` 
2026-04-10 16:56:13.781 zathura[45156:939687] +[NSXPCSharedListener endpointForReply:withListenerName:replyErrorCode:]: an error occurred while attempting to obtain endpoint for listener 'com.apple.view-bridge': Connection invalid
```

Zathura itself works fine, but these errors being spammed to my computer makes it a big hog on cpu resources. 

### Steps to reproduce

Open a PDF with --fork:

```bash
zathura --fork mydocument.pdf
```

Using `&` to background the command, it seems to be fine:
```bash
zathura mydocument.pdf &
```

As soon as movement is done in the zathura window, the terminal gets flooded with the error.

### Expected behavior

No error spamming :) 

### zathura version (zathura --version)

zathura 0.5.11

### girara version (zathura --version)

girara 0.4.5 (runtime: 0.4.5)

### zathura backend

poppler

madskjeldgaard:
When using the `--fork` flag, the cpu usage of zathura goes from 0.5-1.0% in it's normal case without --fork to around 17-20%. So it's pretty intense.

sebastinas:
As always with macOS bugs, this needs somebody that can implement and test fixes on that platform.

PR #932, opened 2026-05-19
Fix: Fixing fork issues on mac with posix

aanthoonyy:
# Original Issue that is being addressed
[907](https://github.com/pwmt/zathura/issues/907)

# Overview
Macs have these things called _Mach Ports_ which makes `fork()` behave slightly different compared to how linux does it. `fork()` duplicates the parent process which includes the mach ports, but the pid of the mach ports are tied to the parents pid, not the new childs pid. So when the child process tries to connect to the macOS window server using the inherited Mach ports, the connection fails because the window server only recognizes the parent's pid. 

This is a common thing on mac, you can get around with by using `fork()` + `exec()` which replaces the process image which ends up spawning a new process with new mach ports with a new pid. The new process would be a copy of the old process, but you'd want to remove the `--fork` tag to stop it from forking again

Another way to do it is using `posix_spawn()`. `posix_spawn()` is just another way to create a new copy process, but it does it without needing to create a process for the goal of just replacing it like you would with `fork()` + `exec()`. This is what my implementation ended up being

# Screenshots
### Before on the original zathura binary (>> zathura --fork pdf.pdf)
image
- Terminal flooded with errors after pressing like j to move around

### After (>> ./build/zathura --fork _plugins paths_ pdf.pdf)
image
- The zathura process is running in the background and the terminal is free to use
- If you append `2>/tmp/z.log` to the command and cat the log file it'll be empty

# Sources
[posix_spawn documentation](https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_spawn.html)
[fork without exec is a mess on mac os](https://developer.apple.com/forums/thread/122691)
[Careful With That Fork, Eugene!](https://developer.apple.com/library/archive/technotes/tn2083/_index.html#//apple_ref/doc/uid/DTS10003794-CH1-SUBSECTION52)


zathura/fork-macos.c
```diff
@@ -0,0 +1,41 @@
+#ifdef __APPLE__
+#include <glib.h>
+#include <unistd.h>
+
+#include "fork-macos.h"
+
+char** build_reexec_argv(char** orig_argv, int orig_argc, char** post_parse_files, int n_files, const char* specific_file) {
```
sebastinas: This does not need to be in its own file. Please move it to main.c.

zathura/main.c
```diff
@@ -190,9 +196,29 @@ GIRARA_VISIBLE int main(int argc, char* argv[]) {
...
   if (print_version == false && argc > file_idx_base + 1) {
-    const pid_t parent_pid              = getpid();
     g_autoptr(girara_list_t) child_pids = girara_list_new();
 
+#ifdef __APPLE__
+    for (int idx = file_idx_base; idx < argc; ++idx) {
+      char** spawn_argv =
+          build_reexec_argv(orig_argv, orig_argc, argv + file_idx_base, argc - file_idx_base, argv[idx]);
+      pid_t pid;
+      const int err = posix_spawn(&pid, spawn_argv[0], NULL, NULL, spawn_argv, environ);
```
sebastinas: So, what is glib doing? If we could rewrite this to use some glib function that does the correct approach internally, we could keep this part generic.

sebastinas (changes requested):

aanthoonyy:
I will finish this tomorrow

zathura/main.c
aanthoonyy: I found [spawn_async_with_pipes_and_fds](https://docs.gtk.org/glib/func.spawn_async_with_pipes_and_fds.html) in the docs. 

The docs say "On modern UNIX platforms, GLib can use an efficient process launching codepath driven internally by posix_spawn()", which seems to be exactly what we want. It removes the need for my Apple specific code branches because glib handles it all for us.

Behavior seems to work as expected 
image

Let me know what you think

aanthoonyy: This might work too, but I'll leave it as is for now
https://docs.gtk.org/glib/func.spawn_async.html

aanthoonyy: I spoke to some friends at work and was told using `g_spawn_async` would be better as it adds a layer of abstraction which reduces the amount of parameters that need to be passed in since we don't need need the control over pipes and fd. So I updated the PR to reflect that. Everything still works as expected

image

sebastinas:
Thanks, I have manually rebased your PR and re-added the call to `setsid`.

Issue #939, opened 2026-05-30
Make #913 link opening confirmation optional


antinomie8:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem

Since #913, a confirmation dialog was added for opening all external links. As I said in https://github.com/pwmt/zathura/pull/913#issuecomment-4472142761, I find it slightly annoying to have to confirm each time I want to open an http link in my web browser for instance, and for my use case the benefit is definitely not worth the cost; from the reactions to this message it seems at least two other persons agree.

### Expected behavior

Would it be possible to add a configuration option for skipping the confirmation dialog altogether when opening external links ?

aanthoonyy:
@sebastinas is this something i can take a stab at?

sebastinas:
Sure, but be aware that the big GTK 4 port is currently in progress. So it might take some time (and might require a rebase) before a PR for this issue can be merged.

aanthoonyy:
No worries regarding that on my end, I'll see it through

sebastinas:
Fixed by #940.

PR #940, opened 2026-06-02
Feature/913 link opening confirmation optional

aanthoonyy:
## Link to original feature request 
[939](https://github.com/pwmt/zathura/issues/939#issuecomment-4595940489)

## Overview
Add a configuration option for skipping the confirmation dialog when opening external links in a document

`:set disable-link-confirm false` is default and keeps behavior the same as before which was providing a dialog that can be accepted by 'y' or 'enter' but anything else cancels opening the link in the browser

 `set disable-link-confirm true` removes the dialogue prompt and just opens the link in the browser

## Changes
- Added the setting via `girara_setting_add` in `config.c`
- Updated `cb_link_confirm` to allow entry to be NULL in `links.c`
- Updated `link_confirm_spawn` to pass entry in as NULL to `cb_link_confirm` in the event that `disable_link_confirm` is true

### DIFF
--- a/zathura/config.c
+++ b/zathura/config.c
@@ -607,6 +607,8 @@ void config_load_default(zathura_t* zathura) {
   girara_setting_add(gsession, "selection-keep-highlight",   &bool_value,  BOOLEAN, false, _("Keep selection highlighted after mouse release"), NULL, NULL);
   bool_value = false;
   girara_setting_add(gsession, "show-signature-information", &bool_value,  BOOLEAN, false, _("Disable additional information for signatures embedded in the document."), cb_show_signature_info, NULL);
+  bool_value = true;
+  girara_setting_add(gsession, "open-link-confirm", &bool_value, BOOLEAN, false, _("Enable prompt to confirm when opening links"), NULL, NULL);
 
   // clang-format on
   // apply some color settings that need their callbacks to run
--- a/zathura/links.c
+++ b/zathura/links.c
@@ -330,10 +330,17 @@ void zathura_link_evaluate(zathura_t* zathura, zathura_link_t* link) {
     girara_debug("Going to remote destination: %s", link->target.value);
     link_confirm(zathura, link->type, link->target.value);
     break;
-  case ZATHURA_LINK_URI:
+  case ZATHURA_LINK_URI: {
+    bool confirm = true;
+    girara_setting_get(zathura->ui.session, "open-link-confirm", &confirm);
     girara_debug("Opening URI: %s", link->target.value);
-    link_confirm(zathura, link->type, link->target.value);
+    if (confirm) {
+      link_confirm(zathura, link->type, link->target.value);
+    } else {
+      link_launch(zathura, link->target.value);
+    }
     break;
+  }
   case ZATHURA_LINK_LAUNCH:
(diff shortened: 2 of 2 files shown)


aanthoonyy:
I found that there are a handful of ways to implement this, so if there is a better way to implement this let me know. 

I will wait for it to get unblocked by the gtk4 port, once that gtk4 port is merged I'll make the needed changes to this branch.

valoq:
- The disable-link-confirm setting turns off the prompt for all external link types, including launch and remote links, not just web links. It should only skip the prompt for URI (web) links and keep confirming launch and remote links.

- cb_link_confirm now treats a NULL entry as "confirmed" and opens the link instead of cancelling. Handle the disabled case in zathura_link_evaluate and leave cb_link_confirm unchanged. The comment above cb_link_confirm is also now wrong.

- The name disable-link-confirm is a double negative and is confusing. a positive name such as link-confirm with default true may be better.

aanthoonyy:
@valoq thanks for the feedback, thoughts on me putting the check in the `ZATHURA_LINK_URI` case?

valoq:
> @valoq thanks for the feedback, thoughts on me putting the check in the `ZATHURA_LINK_URI` case?

Looks ok to me. clang format is complaining though, the spacing is a bit off.

aanthoonyy:
I need to fix my formatter on my config at some point 

I have a question on your comment since I'm relatively new to C 

>     * cb_link_confirm now treats a NULL entry as "confirmed" and opens the link instead of cancelling. Handle the disabled case in zathura_link_evaluate and leave cb_link_confirm unchanged. The comment above cb_link_confirm is also now wrong.
> 

I do react stuff at work, and if you had something in your component that was null, react will just skip over rendering. My thought going in to this was if you pass entry in as null, you'll know that no input is expected to be read and you can just move onto opening your link. Is this idea no relevant to C code?

aanthoonyy:
Also I noticed on your gtk4 pr you combined your commits into one, is that what is expected when committing to zathura?

sebastinas:
There is no clear rule on that. In some cases it makes sense to squash the commits, sometimes not. In this case, as the commits fix up issues in earlier commits of the PR, I would prefer that the commits be squashed. But let's wait with that until we landed the Gtk 4 PR.

aanthoonyy:
Sounds good

sebastinas:
@aanthoonyy  Please rebase on current develop and squash the commits.

aanthoonyy:
@sebastinas Okay, this should be good...

gsadovski:
`set open-link-confirm "false"` works for https links, but not for local path:

2026-08-11-030802_772x524_scrot

aanthoonyy:
I have an idea I'll try when I get off work, thanks for bringing that up

aanthoonyy:
> * The disable-link-confirm setting turns off the prompt for all external link types, including launch and remote links, not just web links. It should only skip the prompt for URI (web) links and keep confirming launch and remote links.

@gsadovski That is actually intentional, but easy to adjust. @sebastinas @valoq what should I do about this?

sebastinas:
I'd make the setting ignore all confirmation requests for external targets.

PR #951, opened 2026-06-18
merge main functions

valoq:
This PR removes the duplicated main function code and merges the sandbox main into main.c
This should not change any functionality for either resulting binary


sebastinas:
SyncTeX support does not work without dbus so there is no point in having the `WITH_SYNCTEX` part enabled if the sandbox binary is built.

PR #952, opened 2026-06-19
Add page-mode setting to zathurarc

### LINKED ISSUE #904: Add page-mode as a configuration option

### Problem

Currently, `page_mode` (which handles `equal_width` and `equal_height` sizing) can only be triggered via shortcut mappings (e.g., `map S page_mode equal_width`). There is no corresponding `set` option in `zathurarc` to define a default behavior upon opening a document.

For users who frequently work with documents containing inconsistent page sizes, it is cumbersome to manually trigger the page mode every time a file is opened.

### Expected behavior

Add a `page-mode` (or `page_mode`) option to the settings. This would allow users to define the default sizing behavior in their `zathurarc`.

```
set page-mode equal_width
```

viv223345:
Add a new 'page-mode' string setting that controls the default page sizing behavior when opening a document. Valid values are 'none' (default), 'equal_width', and 'equal_height'.

Closes #904


sebastinas:
We already have `adjust-open` for that. If the modes are missing, they should be added there.

viv223345:
@sebastinas shall i fold the :page_mode shortcut into :adjust by extending :adjust to accept page_mode arguments?

Issue #953, opened 2026-06-19
Performance Improvements


valoq:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem

There are currently a few architectural issues that impact performance of zathura, especially regarding the startup time. To ensure no noticeable delay, the startup time from command execution to content display should be <0.1 seconds. For small and simple files, this is already fine. For more common and complex file the startup time noticeable in the range of >0.2 seconds and for large files like the 100MB arm spec document linked [here](https://github.com/pwmt/zathura/issues/905) the startup time is multiple seconds

I have looked a bit into the most relevant issues and identified the following three with the most significant impact:
- to re identify files in the database, zathura creates sha256 of all files (even when no database is used). A 50MB file takes 200ms to calculate that hash, a 500MB file needs 2 seconds. This is even before any rendering happens.
We dont need a cryptographic secure hash for this use case, so switching to XXH3-128 from libxxhash would make this >20 times faster. https://github.com/Cyan4973/xxHash

- currently, zathura parses all pages to to lay out the page grid. for a document like the ARM spec pdf with 14734 pages is takes about 2 seconds on modern high end hardware. moving to a lazy approach parsing would fix this. (Effectively meanign that zathura shows the first page immediately, and then continues to parse the following pages)

- before showing anything zathura builds one GtkWidget per page, for the ARM document thats another 500ms. same fix as lazy parsing

Im currently looking into clean solutions that address these, the first one is pretty simple and effectively done. Any feedback and pointers would be much welcome though, especially regarding whether these changes would be supported.

### Expected behavior

zathura should display the documents first page in less then 0.1 seconds regardless of file structure, pages or size

valoq:
Im still working on the background widget loading, but I can already say that the approach works as intended.

Using the Arm spec documentation as example https://developer.arm.com/documentation/ddi0487/lb/ (100MB size, 15k pages):

Currently zathura needs about 2 seconds until the document is opened.
With the background parsing and widget loading, the first page is displayed in 0.05 seconds and can be navigated immediately. For search and index, it takes another 1.2 seconds to execute those features, but that would only be noticeable when the user immediately triggers those features within the first second anyway.

PR #954, opened 2026-06-19
replace sha256 with XXH3-128

valoq:
Currently zathura uses sha256 for file identification in the database, which can take a significant amount of time to calculate for larger files before the parsing even starts. We dont need a cryptographic secure hash for this use case and the alternative XXH3-128 from libxxhash is much faster. This patch also skips the calculation entirely when the db is not used (eg. when running zathura-sandbox)
```

 file size            SHA-256        XXH3-128
 5 MB                 20 ms          0.5 ms
 50 MB                200 ms         8.5 ms
 500 MB               2000 ms        85 ms
```

Partly addresses #953

### DIFF
--- a/zathura/database-sqlite.c
+++ b/zathura/database-sqlite.c
@@ -12,7 +12,7 @@
 #include "utils.h"
 
 /* version of the database layout */
-#define DATABASE_VERSION 3
+#define DATABASE_VERSION 4
 
 static char* sqlite3_column_text_dup(sqlite3_stmt* stmt, int col) {
   return g_strdup((const char*)sqlite3_column_text(stmt, col));
@@ -193,7 +193,7 @@ static void sqlite_db_check_layout(sqlite3* session, const int database_version,
                                           "position_y FLOAT,"
                                           "time TIMESTAMP,"
                                           "page_right_to_left INTEGER,"
-                                          "sha256 BLOB,"
+                                          "hash BLOB,"
                                           "adjust_mode INTEGER"
                                           ");";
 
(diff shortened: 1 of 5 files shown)


zathura/zathura.c
```diff
@@ -1275,12 +1275,14 @@ zathura_fileinfo_t zathura_get_prefileinfo(zathura_t* zathura) {
 static void save_fileinfo_to_db(zathura_t* zathura) {
   zathura_document_t* document = zathura_get_document(zathura);
   const char* path             = zathura_document_get_path(document);
-  const uint8_t* file_hash     = zathura_document_get_hash(document);
 
   zathura_fileinfo_t file_info = zathura_get_fileinfo(zathura);
 
   /* save file info */
-  zathura_db_set_fileinfo(zathura->database, path, file_hash, &file_info);
```
sebastinas, line 1283: There is really no need for this. For the null database, this is a noop.

zathura/document.c
```diff
@@ -228,7 +231,14 @@ const uint8_t* zathura_document_get_hash(zathura_document_t* document) {
     return NULL;
   }
 
-  return document->hash_sha256;
+  if (document->hash_computed == false) {
+    document->hash_computed = true;
```
sebastinas, line 236: I don't remember if instances of `zathura_document_t` are reused on reloads … but in that case this flag might need to be cleared.

valoq: It does not look like they are reused

zathura/database-sqlite.c
```diff
@@ -230,6 +230,9 @@ static void sqlite_db_check_layout(sqlite3* session, const int database_version,
   /* update fileinfo table (part 6) */
   static const char SQL_FILEINFO_ALTER6[] = "ALTER TABLE fileinfo ADD COLUMN sha256 BLOB;";
 
+  /* update fileinfo table (part 7): the column now holds a generic file hash */
+  static const char SQL_FILEINFO_ALTER7[] = "ALTER TABLE fileinfo RENAME COLUMN sha256 TO hash;";
```
sebastinas: The old data is unusable, so I'd prefer removing sha256 and adding hash.

zathura/document.c
```diff
@@ -66,24 +68,28 @@ static bool hash_file_sha256(uint8_t* dst, const char* path) {
...
+  if (success == true) {
+    XXH128_canonical_t canonical;
+    XXH128_canonicalFromHash(&canonical, XXH3_128bits_digest(state));
+    memcpy(dst, canonical.digest, DIGEST_SIZE);
   }
 
-  gsize dst_size = DIGEST_SIZE;
-  g_checksum_get_digest(checksum, dst, &dst_size);
-  return true;
+  XXH3_freeState(state);
```
sebastinas: The restructuring here could be avoided by adding a `G_DEFINE_AUTOPTR_CLEANUP_FUNC(XXH3_state_t, XXH3_freeState)`.

PR #955, opened 2026-06-19
Adding setting to start in single page mode

aanthoonyy:
## link to original post 
[904](https://github.com/pwmt/zathura/issues/904#issuecomment-4750212994)

Adding `set single-page-mode true/false` to your config lets you toggle on single page mode on start up

### DIFF
--- a/zathura/config.c
+++ b/zathura/config.c
@@ -609,6 +609,8 @@ void config_load_default(zathura_t* zathura) {
   girara_setting_add(gsession, "show-signature-information", &bool_value,  BOOLEAN, false, _("Disable additional information for signatures embedded in the document."), cb_show_signature_info, NULL);
   bool_value = true;
   girara_setting_add(gsession, "open-link-confirm",          &bool_value,  BOOLEAN, false, _("Enable prompt to confirm when opening links"), NULL, NULL);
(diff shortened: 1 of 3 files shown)


aanthoonyy:
@sebastinas do you want this to have a callback

sebastinas:
There is no need for a callback. When creating the document widget, query the setting and set it accordingly. For toggling the view setting in an open document, we already have a shortcut.

aanthoonyy:
Sounds good, I'll update the PR tonight

aanthoonyy:
dont merge yet, give me a couple hours to come back

aanthoonyy:
Ok this PR is good to go. I just saw your comment in #961, if you prefer the other guys implementation over this one feel free to use that. Although, it does have a bug where when you toggle off the single page mode with your shortcut it goes to the top of the page.

aanthoonyy:
Also, when coming out of single page mode using the shortcut, there seems to be a long load time for larger documents (I'm using the "the art of unix programing" as a test). The load time stops you from being able to scroll through pages for like 10 seconds. 

I'm moving this week so I don't have too much more time to investigate this until after I'm settled in at my new place but wanted to bring it to your attention for the time being.

PR #956, opened 2026-06-21
perf: implement background parsing

valoq:
Partly addresses https://github.com/pwmt/zathura/issues/953


valoq:
I still need to look into some issues for this one and do some more testing

PR #958, opened 2026-06-22
Improve quit action

valoq:
Currently, zathura handles the quit action the same as the close document action.
However, that results in a lot of memory cleanup that is only needed when a document is closed while keeping the process.
As a consequence, pressing quit on a large document, can cause a delay of about a second until zathura actually closes.
If I didnt overlook anything here, this seems easy to avoid by using a dedicated quit path that skips the cleanup since the process dies anyway. With this patch, zathura quits immediately as expected

### DIFF
--- a/zathura/callbacks.c
+++ b/zathura/callbacks.c
@@ -27,22 +27,9 @@
 #include "synctex.h"
 #include "dbus-interface.h"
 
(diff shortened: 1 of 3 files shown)


sebastinas:
No, we will do proper clean up. This is otherwise a pain to debug memory leaks.

PR #961, opened 2026-06-24
Add single-page-mode bool setting, for starting in that mode

### LINKED ISSUE #904: Add page-mode as a configuration option

quite:
Closes https://github.com/pwmt/zathura/issues/904 (what's left to do there)

### DIFF
--- a/doc/man/zathurarc.5.rst
+++ b/doc/man/zathurarc.5.rst
@@ -961,6 +961,12 @@ The settings described here can be changed with ``set``.
   * Value type: String
   * Default value: rgba(100%,84%,0%,0.9)
 
+*single-page-mode*
+  Start in single page display mode (not page grid)
+
+  * Value type: Boolean
+  * Default value: false
+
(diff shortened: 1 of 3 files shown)


sebastinas:
There is already #955.

quite:
Oh sorry I missed that one. It does look somewhat convoluted currently, we'll see.

quite:
Closing in favour of #955

PR #962, opened 2026-06-24
implement lazy parsing

valoq:
Instead of parsing all pages before rendering the first page, this now parses only the displayed pages and everything else on demand. parsing is relatively fast but it adds a delay if 15k pages are all parsed before the document is shown. The lazy parsing seems to be fine for the subsequent pages as I could not find any relevant performance impact. In theory we could make the other pages preload the parsing but that would add complexity for (as far as I see) nothing measurable.

There is one relevant side effect:
Previously the layout was build based on the size of all pages in the document, which isn't possible with lazy parsing. While this isn't relevant for most files, it changes the behavior for files with pages of different layout/size and automatically changes the layout once a page with a different layout from the first page(s) is parsed and rendered.

Partly addresses #953

### DIFF
--- a/zathura/commands.c
+++ b/zathura/commands.c
@@ -436,9 +436,7 @@ bool cmd_search(girara_session_t* session, const char* input, girara_argument_t*
     GObject* obj_page_widget = G_OBJECT(page_widget);
     g_object_set(obj_page_widget, "draw-links", FALSE, NULL);
 
-    zathura_renderer_lock(zathura->sync.render_thread);
     girara_list_t* result = zathura_page_search_text(page, input, &error);
-    zathura_renderer_unlock(zathura->sync.render_thread);
 
     if (result == NULL || girara_list_size(result) == 0) {
       girara_list_free(result);
--- a/zathura/completion.c
+++ b/zathura/completion.c
@@ -20,6 +20,8 @@
 #include "utils.h"
 #include "page.h"
 #include "database.h"
+#include "render.h"
+#include "zathura.h"
 
 static int compare_case_insensitive(const void* data1, const void* data2) {
   const char* str1 = data1;
--- a/zathura/page-widget.c
+++ b/zathura/page-widget.c
@@ -13,6 +13,7 @@
 
 #include "links-internal.h"
 #include "page.h"
+#include "adjustment.h"
 #include "render.h"
 #include "utils.h"
 #include "shortcuts.h"
@@ -877,6 +878,33 @@ static void cb_update_surface(ZathuraRenderRequest* UNUSED(request), cairo_surfa
   ZathuraPageWidget* widget = data;
   g_return_if_fail(ZATHURA_IS_PAGE_WIDGET(widget));
   zathura_page_widget_update_surface(widget, surface, false);
+
+  if (surface == NULL) {
+    return;
+  }
+
+  /* the page is now parsed, correct its size if it differs from the placeholder */
+  ZathuraPageWidgetPrivate* priv = zathura_page_widget_get_instance_private(widget);
+  zathura_document_t* document   = zathura_page_get_document(priv->page);
+  unsigned int page_width = 0, page_height = 0;
+  page_calc_height_width(document, priv->page, &page_height, &page_width, true);
+
+  int cur_width = 0, cur_height = 0;
+  gtk_widget_get_size_request(GTK_WIDGET(widget), &cur_width, &cur_height);
+  if ((int)page_width != cur_width || (int)page_height != cur_height) {
+    /* resize directly to keep the fresh surface, it was already rendered at the corrected size */
+    gtk_widget_set_size_request(GTK_WIDGET(widget), (int)page_width, (int)page_height);
+    if (priv->drawing_area != NULL) {
+      gtk_widget_set_size_request(priv->drawing_area, (int)page_width, (int)page_height);
+      gtk_widget_queue_draw(priv->drawing_area);
+    }
+  }
+
+  /* refresh the statusbar so the label of a freshly parsed current page shows up */
+  if (zathura_page_label_is_number(priv->page) == false && zathura_page_get_label(priv->page, NULL) != NULL &&
+      zathura_page_get_index(priv->page) == zathura_document_get_current_page_number(document)) {
+    statusbar_page_number_update(priv->zathura);
+  }
 }
 
 static void cb_cache_added(ZathuraRenderRequest* UNUSED(request), void* data) {
--- a/zathura/page.c
+++ b/zathura/page.c
@@ -23,6 +23,7 @@ struct zathura_page_s {
   unsigned int index;           /**< Page number */
   bool visible;                 /**< Page is visible */
   bool label_is_number;         /**< Page label is the same as the page number */
+  gint loaded;                  /**< Page has been parsed by the plugin, atomic */
 };
 
 zathura_page_t* zathura_page_new(zathura_document_t* document, unsigned int index, zathura_error_t* error) {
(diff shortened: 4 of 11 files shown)


aanthoonyy:
Curious if this would help improve this 'issue' I found 
https://github.com/pwmt/zathura/pull/955#issuecomment-4795165176

sebastinas:
FWIW, I'd like to get the Gtk 4 related regressions addressed first before looking at this PR and #964 .

zathura/page.c
```diff
@@ -47,44 +48,50 @@ zathura_page_t* zathura_page_new(zathura_document_t* document, unsigned int inde
...
     ret = functions->page_get_label(page, page->data, &page->label);
     if (ret != ZATHURA_ERROR_OK) {
-      if (error != NULL) {
-        *error = ret;
-      }
-      goto error_free;
-    }
-
-    if (page->label != NULL) {
+      girara_warning("Failed to get label of page %u: %d", page->index + 1, ret);
```
sebastinas: This does not deserve a warning. This is info at most.

zathura/page.c
```diff
@@ -47,44 +48,50 @@ zathura_page_t* zathura_page_new(zathura_document_t* document, unsigned int inde
...
+    girara_error("Failed to initialize page %u: %d", page->index + 1, ret);
+    return false;
   }
 
-  /* get label if there is one */
+  /* a missing label is not fatal, the page number is shown instead */
   if (functions->page_get_label != NULL) {
     ret = functions->page_get_label(page, page->data, &page->label);
     if (ret != ZATHURA_ERROR_OK) {
-      if (error != NULL) {
```
sebastinas: Why are you removing this? If the plugin function returns an actual error, the error is to be propagated to the caller.

sebastinas:
I think this PR misuses the renderer lock as document lock. It would be cleaner to move the lock to `zathura_docment_t` if calls to document/page functions no need to be locked.

zathura/page.c
valoq: zathura_page_load now returns bool and no longer has an error parameter which means there is no caller to propagate the label error to. 
Should a failed page_get_label keep loading the page without a label as it does now or fail the page load?

sebastinas: No. Then `zathura_page_load` needs to propagate the error. You are moving part of the functionality of `zathura_page_new` to the new function, then please also move the error handling. A `bool` is not sufficient in this case. (Admittedly, we are not good at handling the error, but he common approach is: functions in the document and page API, propagate the errors from the plugins).  In this new function would part of that API.

zathura/commands.c
```diff
@@ -437,7 +437,9 @@ bool cmd_search(girara_session_t* session, const char* input, girara_argument_t*
     g_object_set(obj_page_widget, "draw-links", FALSE, NULL);
 
     zathura_renderer_lock(zathura->sync.render_thread);
```
sebastinas: This should now be a document lock.

zathura/commands.c
```diff
@@ -437,7 +437,9 @@ bool cmd_search(girara_session_t* session, const char* input, girara_argument_t*
...
-    girara_list_t* result = zathura_page_search_text(page, input, &error);
+    /* parse on demand so search covers pages not parsed yet */
+    girara_list_t* result =
+        zathura_page_load(page, &error) == true ? zathura_page_search_text(page, input, &error) : NULL;
```
sebastinas: For all these changes that are now sprinkled throughout the code base, I'd prefer that the `zathura_page_*` takes care of a missing load.

zathura/commands.c
```diff
@@ -542,6 +544,11 @@ bool cmd_export(girara_session_t* session, girara_list_t* argument_list) {
       goto image_error;
     }
 
+    /* parse on demand so images of pages not parsed yet can be exported */
+    if (zathura_renderer_load_page(zathura->sync.render_thread, page) == false) {
+      goto image_error;
+    }
+
```
sebastinas: As above

zathura/completion.c
```diff
@@ -340,6 +342,11 @@ girara_completion_t* cc_export(girara_session_t* session, const char* input) {
       continue;
     }
 
+    /* parse on demand so images of pages not parsed yet are listed */
+    if (zathura_renderer_load_page(zathura->sync.render_thread, page) == false) {
```
sebastinas: As above

zathura/page-widget.c
```diff
@@ -429,7 +439,8 @@ static void zathura_page_widget_set_property(GObject* object, guint prop_id, con
     priv->signatures.draw = g_value_get_boolean(value);
     /* get links */
     if (priv->signatures.draw == TRUE && priv->signatures.retrieved == FALSE) {
-      priv->signatures.list      = zathura_page_get_signatures(priv->page, NULL);
+      priv->signatures.list =
+          page_widget_load_page(priv) == true ? zathura_page_get_signatures(priv->page, NULL) : NULL;
```
sebastinas: As above

zathura/page-widget.c
```diff
@@ -359,7 +369,7 @@ static void zathura_page_widget_set_property(GObject* object, guint prop_id, con
     priv->links.draw = g_value_get_boolean(value);
     /* get links */
     if (priv->links.draw == TRUE && priv->links.retrieved == FALSE) {
-      priv->links.list      = zathura_page_links_get(priv->page, NULL);
+      priv->links.list      = page_widget_load_page(priv) == true ? zathura_page_links_get(priv->page, NULL) : NULL;
```
sebastinas: As above

zathura/page-widget.c
```diff
@@ -906,7 +944,7 @@ static void evaluate_link_at_mouse_position(ZathuraPageWidget* page, int oldx, i
   /* simple single click */
   /* get links */
   if (priv->links.retrieved == false) {
-    priv->links.list      = zathura_page_links_get(priv->page, NULL);
+    priv->links.list      = page_widget_load_page(priv) == true ? zathura_page_links_get(priv->page, NULL) : NULL;
```
sebastinas: As above

zathura/page-widget.c
```diff
@@ -1073,7 +1111,7 @@ static void cb_zathura_page_widget_button_release_event(GtkGestureClick* gesture
     tmp.y1 /= scale;
     tmp.y2 /= scale;
 
-    g_autofree char* text = zathura_page_get_text(priv->page, tmp, NULL);
+    g_autofree char* text = page_widget_load_page(priv) == true ? zathura_page_get_text(priv->page, tmp, NULL) : NULL;
```
sebastinas: As above

zathura/page-widget.c
```diff
@@ -1145,15 +1183,16 @@ static void cb_zathura_page_widget_motion_notify(GtkEventControllerMotion* contr
       selection.x2 /= scale;
       selection.y2 /= scale;
 
-      priv->selection.list = zathura_page_get_selection(priv->page, selection, NULL);
+      priv->selection.list =
+          page_widget_load_page(priv) == true ? zathura_page_get_selection(priv->page, selection, NULL) : NULL;
```
sebastinas: As above

zathura/page-widget.c
```diff
@@ -1145,15 +1183,16 @@ static void cb_zathura_page_widget_motion_notify(GtkEventControllerMotion* contr
...
       if (priv->selection.list != NULL && girara_list_size(priv->selection.list) != 0) {
         priv->selection.draw = true;
         zathura_page_widget_redraw_canvas(page);
       }
     }
   } else {
     if (priv->links.retrieved == false) {
-      priv->links.list      = zathura_page_links_get(priv->page, NULL);
+      priv->links.list      = page_widget_load_page(priv) == true ? zathura_page_links_get(priv->page, NULL) : NULL;
```
sebastinas: As above

zathura/page-widget.c
```diff
@@ -1206,7 +1245,7 @@ static void zathura_page_widget_popup_menu(GtkWidget* widget, double x, double y
   ZathuraPageWidgetPrivate* priv = zathura_page_widget_get_instance_private(page);
 
   if (priv->images.retrieved == false) {
-    priv->images.list      = zathura_page_images_get(priv->page, NULL);
+    priv->images.list      = page_widget_load_page(priv) == true ? zathura_page_images_get(priv->page, NULL) : NULL;
```
sebastinas: As above

zathura/print.c
```diff
@@ -25,13 +26,20 @@ static bool draw_page_cairo(cairo_t* cairo, zathura_t* zathura, zathura_page_t*
   /* Try to render the page without a temporary surface. This only works with
    * plugins that support rendering to any surface.  */
   zathura_renderer_lock(zathura->sync.render_thread);
-  const int err = zathura_page_render(page, cairo, true);
+  /* parse on demand so pages not parsed yet can be printed */
+  const int err =
+      zathura_page_load(page, NULL) == true ? zathura_page_render(page, cairo, true) : ZATHURA_ERROR_UNKNOWN;
```
sebastinas: As above

zathura/print.c
```diff
@@ -25,13 +26,20 @@ static bool draw_page_cairo(cairo_t* cairo, zathura_t* zathura, zathura_page_t*
...
   zathura_renderer_unlock(zathura->sync.render_thread);
 
   return err == ZATHURA_ERROR_OK;
 }
 
 static bool draw_page_image(cairo_t* cairo, GtkPrintContext* context, zathura_t* zathura, zathura_page_t* page) {
+  /* parse on demand so the real page size is used for the temporary surface */
+  if (zathura_renderer_load_page(zathura->sync.render_thread, page) == false) {
```
sebastinas: If we reach that, why is the page not loaded yet?

zathura/print.c
```diff
@@ -120,8 +128,13 @@ static void cb_print_request_page_setup(GtkPrintOperation* UNUSED(print_operatio
   }
 
   zathura_page_t* page = zathura_document_get_page(zathura_get_document(zathura), page_number);
-  double width         = zathura_page_get_width(page);
-  double height        = zathura_page_get_height(page);
+  /* parse on demand so the orientation matches the real page size */
+  if (page == NULL || zathura_renderer_load_page(zathura->sync.render_thread, page) == false) {
```
sebastinas: We really should not be able to have unloaded pages here. If loading pages failed, we really shouldn't be printing.

zathura/commands.c
valoq: the load uses the document lock now and the renderer lock only guards rendering

zathura/print.c
valoq: yes, the check is now removed

zathura/print.c
valoq: dropped the load guard and the render call reports the error instead

sebastinas:
Okay, let's go ahead with this. @valoq could you rebase it so that I can merge it?

PR #964, opened 2026-06-25
Create page widgets in the background

valoq:
With this PR only the on screen page widget is created when a document is opened and the rest are created in the background at low priority, so opening documents with many pages are displayed immediately instead of taking seconds.
There is still on demand widget creation, so immediate scrolling still works as well

together with the lazy parsing PR, this reduces the display time of large documents from multiple seconds to <0.1 second

### DIFF
--- a/zathura/zathura.c
+++ b/zathura/zathura.c
@@ -825,6 +825,165 @@ char* get_formatted_filename(zathura_t* zathura, bool statusbar) {
   }
 }
 
+/* create the page widget if missing, then attach it to the grid */
+void create_page_widget(zathura_t* zathura, unsigned int page_id) {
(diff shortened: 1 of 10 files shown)


zathura/zathura.c
```diff
@@ -1273,6 +1446,11 @@ zathura_fileinfo_t zathura_get_prefileinfo(zathura_t* zathura) {
 }
 
 static void save_fileinfo_to_db(zathura_t* zathura) {
+  if (ZATHURA_IS_NULLDATABASE(zathura->database)) {
```
sebastinas: This is unrelated, please remove.

sebastinas:
Hi @valoq. I haven't had time to look at this PR yet, but could you rebase it so that it would potentially be mergeable?

PR #975, opened 2026-07-08
Fix scroll sensitivity

### LINKED ISSUE #974: Extremely high scroll sensitivity since the update to gtk4

### Problem

As the title says, since the update to the latest release on Arch the scroll sensitivity of my laptop touchpad basically skyrocketed, making it almost unusable. I suspect it has something to do with the migration to gtk4, but other gtk4 apps (like rnote) seem to work fine.

I'm not entirely sure if this is a problem in zathura itself, but in other apps where I usually scroll with my touchpad a lot (firefox, xournalpp and the aforementioned rnote) I never encountered this issue before.

### Steps to reproduce

1. Open a pdf
2. Scroll with the touchpad

### Expected behavior

Touchpad scrolling behaves as in other apps

2026.07.02

2026.02.04

### zathura backend

mupdf (the same problem arises with the poppler backend)

valoq:
GTK4 sends touchpad deltas as raw pixels, but the code still expects GTK3 deltas which were 10 times smaller.

Fixes https://github.com/pwmt/zathura/issues/974

### DIFF
--- a/girara-gtk/callbacks.c
+++ b/girara-gtk/callbacks.c
@@ -358,13 +358,15 @@ gboolean girara_callback_view_scroll_event(GtkEventControllerScroll* controller,
   /* only wheel units are discrete steps so touchpad deltas stay smooth */
   if (gtk_event_controller_scroll_get_unit(controller) != GDK_SCROLL_UNIT_WHEEL) {
     event.type = GIRARA_EVENT_SCROLL_BIDIRECTIONAL;
-    event.x    = dx;
-    event.y    = dy;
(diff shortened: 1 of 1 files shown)


sebastinas:
Where is 10 coming from?

valoq:
> Where is 10 coming from?

In GTK3 the toolkit divided touchpad scroll values by 10 before passing them to the app. GTK4 doenst and passes the raw values, which makes scrolling become too fast. The patch just divides the values by 10 to restore the gtk3 previous.

PR #989, opened 2026-07-15
fix: crash when open doc fails

valoq:
The error paths near the top of document_open jump over the declarations of first_page_column_list and page_mode. But both are freed when the function returns and free is called on uninitialized values whenever one of those paths is taken.

I ran into this when I played around with the ABI versions and this triggers when there is a version mismatch, leading to double free.


sebastinas:
There are more cases that still have some goto-based error handling which is mixed with `g_auto`-foo. Implemented a15190e1364459f65a81f18498f446153a299170 in favor to this PR.

PR #1010, opened 2026-08-17
Fix horizontal scrolling

### LINKED ISSUE #983: Horizontal position stuck at the left edge

### Problem

Zooming out to a certain level and zooming back in causes the horizontal position to be stuck and I cannot move left/right (neither with the middle button of the mouse, nor with h/l), the left edge of the page is stuck at the left edge of the window. With the double-page mode, the critical zoom level for me is the one where the content width becomes smaller than the window width, for single-page mode, for single-page, it’s a bit further.
The bug is present with both pdf-mupdf and djvu plugins, but disappears when enabling the horizontal scrollbar (`set guioptions "sh"`).
I’ve recently updated my packages and discovered the bug shortly after it.
Zathura version: 2026.07.02-1
OS, desktop: Arch Linux, Hyprland 0.55.4 (Wayland)

### Steps to reproduce

- Ensure that your configuration doesn’t contain `guioptions`.
- Open any document in Zathura.
- Zoom out as much as you can.
- Zoom back in.

As soon as the content is wider than the window, the displayed region will be stuck to the left edge of the document. No horizontal movement will be possible.

### Expected behavior

I would expect the content to be zoomed towards the cursor or towards the center, not the left edge. It should be possible to view the part hidden beyond the right window edge.

zathura 2026.07.02

girara 2026.02.04 (runtime: 2026.02.04)

### zathura backend

pdf-mupdf, djvu

valoq:
Fixes #983

### DIFF
--- a/zathura/adjustment.c
+++ b/zathura/adjustment.c
@@ -173,19 +173,34 @@ gdouble zathura_adjustment_get_ratio(GtkAdjustment* adjustment) {
   return (value - lower + page_size / 2.0) / (upper - lower);
 }
 
-void zathura_adjustment_set_value(GtkAdjustment* adjustment, gdouble value) {
+/* keep the value inside the scrollable range */
+static gdouble clamp_to_range(GtkAdjustment* adjustment, gdouble value) {
   const gdouble lower        = gtk_adjustment_get_lower(adjustment);
   const gdouble upper_m_size = gtk_adjustment_get_upper(adjustment) - gtk_adjustment_get_page_size(adjustment);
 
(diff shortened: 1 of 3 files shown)


dexgs:
This fixes horizontal scrolling with the keyboard for me, but when I middle click and drag, the document can still only be moved up and down, not horizontally.

valoq:
> This fixes horizontal scrolling with the keyboard for me, but when I middle click and drag, the document can still only be moved up and down, not horizontally.

@dexgs can you confirm if #1011 fixes the issue completely for you?

dexgs:
> can you confirm if #1011 fixes the issue completely for you?

@valoq It seems to, thanks.

PR #1015, opened 2026-08-25
Fix touchpad scroll speed

### LINKED ISSUE #991: Scrolling with touchpad randomly jumps pages

### Problem

While scrolling with the touchpad, zathura randomly jumps opposite to the scrolling direction. Here by "jump" I mean an instantaneous change in position within the document.
For instance, I was scrolling downwards and zathura jumped from page 10 back up to page 2.
Sometimes it goes back to the sart of the page, sometimes by only one page, and sometimes by multiple pages. There doesn't seem to be any noticeable pattern in this jumping bug. No specific page numbers, no specific distance jumped.
This bug manifested only after the GTK4 update. Incidentally, but maybe unrelated, touchpad scrolling also feels janky and inconsistent after the update to GTK4. The last release with good touchpad scrolling was 2026.05.20.
Scrolling with the arrow keys works fine.

### Steps to reproduce

1. Open a PDF document
2. Start scrolling downwards with the touchpad
3. Continue scrolling until zathura jumps up a few pages
OR
2. Start scrolling upwards with the touchpad
3. Continue scrolling untill zathura jumps down a few pages (usualy only 1 page)

### Expected behavior

Scrolling with the touchpad should be smooth and without random backwards jumps.

zathura 2026.07.08

girara 2026.07.07 (runtime: 2026.07.07)

### zathura backend

(plugin) pdf-poppler (2026.5.10) (/usr/lib/zathura/libpdf-poppler.so)

valoq:
Probably fixes #991 (dont have a touchpad to test this on at hand)

currently typing a number will use that number as a factor for scroll speed. I suspect this is an undocumented feature so I left this for the mouse. but for the touch pad the multiplication happened with every registered input, which made the increase much larger.

### DIFF
--- a/girara-gtk/callbacks.c
+++ b/girara-gtk/callbacks.c
@@ -382,7 +382,8 @@ gboolean girara_callback_view_scroll_event(GtkEventControllerScroll* controller,
 #endif
 
   /* only wheel units are discrete steps so touchpad deltas stay smooth */
-  if (gtk_event_controller_scroll_get_unit(controller) != GDK_SCROLL_UNIT_WHEEL) {
+  const bool discrete = gtk_event_controller_scroll_get_unit(controller) == GDK_SCROLL_UNIT_WHEEL;
+  if (discrete == false) {
     event.type = GIRARA_EVENT_SCROLL_BIDIRECTIONAL;
     event.x    = dx * surface_to_wheel;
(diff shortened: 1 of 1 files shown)


iCaird:
Can confirm that this seems to have fixed #991 for me

PR #1020, opened 2026-08-31
Add window-decoration setting to disable CSD on Wayland

### LINKED ISSUE #1018: Add window-decoration setting to allow disabling window decorations (CSD) on Wayland
## Summary

zathura links GTK4, which on Wayland always renders client-side decorations (CSD) and never negotiates the xdg-decoration protocol. As a result the window always has a title bar and there is currently no way to launch it borderless/natively under a Wayland compositor that can't strip CSD (e.g. COSMIC). GTK3's `GTK_CSD` is not relevant for GTK4, and `GDK_BACKEND=x11` still renders CSD.

## Proposal

Add an init-only `window-decoration` boolean setting (default `true`) so users can set:

```
set window-decoration false
```

which calls `gtk_window_set_decorated()` before the window is shown. Default stays `true`, so existing behavior is unchanged.

Related PR: #<pr>
See https://github.com/pwmt/zathura/issues/828

AuthenticSm1les:
Closes #1018

Adds an init-only `window-decoration` boolean setting (default `true`):

- `zathura/config.c`: register the setting (BOOLEAN, init-only, default true)
- `girara-gtk/session.c`: read it and call `gtk_window_set_decorated()` before the window is shown

This lets users run fully borderless under native Wayland (GTK4 renders CSD that a compositor such as COSMIC can't strip):

```
set window-decoration false
```

Default stays `true`, so no behavior change for existing users.

### DIFF
--- a/girara-gtk/session.c
+++ b/girara-gtk/session.c
@@ -418,6 +418,11 @@ bool girara_session_init(girara_session_t* session, const char* sessionname) {
     gtk_window_set_default_size(GTK_WINDOW(session->gtk.window), window_width, window_height);
   }
 
+  /* apply the window-decoration setting (read from the application's config) before showing the window */
+  bool window_decoration = true;
+  girara_setting_get(session, "window-decoration", &window_decoration);
+  gtk_window_set_decorated(GTK_WINDOW(session->gtk.window), window_decoration);
+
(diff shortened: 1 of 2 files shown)


AuthenticSm1les:
Closing per author request.

Issue #1028, opened 2026-09-10
Scroll jumps back ~2 pages intermittently on Wayland (not on X11)


fregas:
zathura 2026.07.18, girara 2026.07.18, Fedora, Wayland session.

While scrolling down (touchpad), the view intermittently jumps back ~2
pages, then continues normally.

Ruled out:
- Backend: reproduces identically with mupdf-only and poppler-only
- Kinetic scrolling: GDK_ENABLE_KINETIC_SCROLLING=0 doesn't fix it
- Compositor/libinput-wide: another GTK PDF viewer (Evince) on the same
  session doesn't reproduce it
- Not reproducible under XWayland (GDK_BACKEND=x11)

Seems isolated to zathura on Wayland specifically.

kirasingla:
I am experiencing the same bug on Arch with zathura and girara 2026.07.18-1 using Sway. The issue does not happen when using the keyboard for window movements, and seems to happen at random as described.

pony-montana:
experiencing the same bug, on arch hyprland desktop

florencebiree:
same here, zathura 2026.07.18-1 from debian, with wayland/sway

sebastinas:
Please test #1030

JimnwQ:
Here's some ways to more easily reproduce the bug on 2026.07.18:
- Open large document (> 400 pages). Go 100 page down (100J). Try spamming short swipe gesture and `-`, `=` key at the same time. (So it scrolls a little bit, quickly stops and zooms out or in, quickly scrolls a little bit again, and so on)
- Try Alt + Touchpad Scrolling and while it's kinetic scrolling, spam `-` key.

My patch seem to solve both of those issues. Would very much appreciate if anyone can test it on large document for long time.

kirasingla:
The patch seems to fix the issues on my end as well. Tested with a 1500 page PDF and things ran smoothly.

sebastinas:
Fixed by #1030.

PR #1030, opened 2026-09-21
Fix cb_view_adjustment_changed* callbacks issues

JimnwQ:
Removes three FIXMEs in cb_view_adjustment_changed*:
- Fix stored_position hacks on cb_view_adjustment_value_changed
- Fix cb_view_adjustment_changed modifying the adjustment value

Closes:
- https://github.com/pwmt/zathura/issues/1028

The random scroll jumpings are occuring because GDK_SCROLL events are handled by GtkScrolledWindow internal scroll event listener, which sets adjustment values to it's private priv->unclamped_vadj_value + scroll_delta. As far as I can understand the girara session's GtkEventControllerScroll should handle all of them and no event should leak into GtkScrolledWindow's event controller, but it's still happening and I don't really know why.

To accommodate it, it's possible to allow both GtkScrolledWindow internal logic and installed mouse shortcuts to modify the adjustments. priv->unclamped_vadj_value and zathura->position_* should be set in adjustment value-changed signal callback, and refresh-view signal writes to the adjustment, so in theory everything is fine. The caveat is, GtkScrolledWindow's logic is bit weird and priv->unclamped_vadj_value is only set when the widget is not kinetic scrolling. And when the GDK_SCROLL event leaks, it starts kinetic scrolling hence never really sets priv->unclamped_vadj_value and just jumps to last place priv->unclamped_vadj_value is set to.

So the fix is simply cancelling kinetic scrolling whenever manually scrolling with position_set, sc_zoom so adjustment value-changed callback can correctly modify priv->unclamped_vadj_value.

This also restores the unintended feature of holding alt and scrolling allowing kinetic scrolling because ALT_MASK + GDK_SCROLL event handler is not installed, falling back to GtkScrolledWindow.

I thoroughly tested by pressing random keybindings on my machine, and so far didn't find any bug.

P.S. Thank you so much for creating such a great document viewer. Really been using zathura as the only document viewer forever.

Void Linux, Kernel 7.2.5, Sway (GDK_BACKEND=wayland)
GTK Version: 4.22.4
Girara Version: 2026.07.18

Edit: The only place where upper bound of adjustments is changed seem to be zooming, hence just restore the position ratio after setting the bounds on zoom.


zathura/shortcuts.c
```diff
@@ -1592,6 +1592,9 @@ bool sc_zoom(girara_session_t* session, girara_argument_t* argument, girara_even
   g_return_val_if_fail(argument != NULL, false);
   g_return_val_if_fail(zathura_has_document(zathura), false);
 
+  gtk_scrolled_window_set_kinetic_scrolling(GTK_SCROLLED_WINDOW(zathura->ui.view), FALSE);
```
sebastinas, line 1600: Please add a comment here why this is necessary.

zathura/zathura.c
```diff
@@ -1653,6 +1653,9 @@ bool position_set(zathura_t* zathura, double position_x, double position_y) {
     return true;
   }
 
+  gtk_scrolled_window_set_kinetic_scrolling(GTK_SCROLLED_WINDOW(zathura->ui.view), FALSE);
```
sebastinas, line 1661: As above.

sebastinas (changes requested):

sebastinas:
It'd appreciate somebody with Wayland to test this PR.

zathura/shortcuts.c
JimnwQ: Done! 861928de510b27fa29ec1520b11588ba79bca76c

